PR #2, opened 2024-10-31
Fix misleading number of bytes sent reported on Apple platforms

jjanowsk:
The number of bytes sent reported is incorrect, this PR fixes it


gytsto (approved):
+2

gytsto (approved):

PR #3, opened 2024-11-28
Only mark timers to update

Hasan6979:
Timers are updated on every single packet call, but their accuracy is limited to 250ms where currentTime is updated. So just mark timers to update, and update them only once in update_timers.

This PR is part 1 of several PR's to revamp the neptun noise logic. The next PR would remove the lock from the peer.tunnel and introduce interior locks to tunnel members instead. In the hotpath, we would only be reading the session so no need to lock the entire tunnel for that. Other members of tunnel are either lockable or can be made atomic. Locking timers internally is not possible, otherwise it can lead to a deadlock as `update_timers` also tries to lock() the timers then.

### DIFF
--- a/neptun/src/noise/timers.rs
+++ b/neptun/src/noise/timers.rs
@@ -420,6 +420,11 @@ impl Tunn {
 
 #[cfg(test)]
 mod tests {
+    use std::{
+        sync::{Arc, Barrier},
+        thread,
+    };
+
     use rand::RngCore;
     use rand_core::OsRng;
 
@@ -467,5 +472,47 @@ mod tests {
         assert!(my_tun.timers[TimerName::TimeLastDataPacketSent].is_zero());
         assert!(my_tun.timers[TimerName::TimeLastDataPacketReceived].is_zero());
         assert!(my_tun.timers[TimerName::TimePersistentKeepalive].is_zero());
+    }
+
+    #[test]
+    fn test_mark_timers_during_update() {
+        let my_secret_key = x25519_dalek::StaticSecret::random_from_rng(OsRng);
+        let my_idx = OsRng.next_u32();
+
+        let their_secret_key = x25519_dalek::StaticSecret::random_from_rng(OsRng);
+        let their_public_key = x25519_dalek::PublicKey::from(&their_secret_key);
+
+        let mut tun = Tunn::new(my_secret_key, their_public_key, None, None, my_idx, None).unwrap();
+
+        let last_data_sent_time = tun.timers[TimerName::TimeLastDataPacketSent];
+
+        let n = 1;
+        let mut handles = Vec::with_capacity(n);
+        let t = Arc::new(tun);
+        let barrier = Arc::new(Barrier::new(n));
+
+        let b = Arc::clone(&barrier);
+        handles.push(thread::spawn(move || {
+            b.wait();
+            t.mark_timer_to_update(TimerName::TimeLastDataPacketSent);
+        }));
+
+        // Trying to marking timer from one thread
+        // And updating the timers in the other
+        // This wouldn't work because tunnel needs
+        // to be mutable, and wrapping a mutex on it
+        // defeats the purpose. Atm, not sure how to test
+        // that code path.
+
+        // let t_clone = t.clone();
+        // let b = Arc::clone(&barrier);
+        // let mut dst = [0u8; 1400];
+        // b.wait();
+        // tun.update_timers(&mut dst);
+
+        // Wait for other threads to finish.
+        for handle in handles {
+            handle.join().unwrap();
+        }
     }
 }
--- a/neptun/src/noise/timers.rs
+++ b/neptun/src/noise/timers.rs
@@ -6,6 +6,7 @@ use super::errors::WireGuardError;
 use crate::noise::{safe_duration::SafeDuration as Duration, Tunn, TunnResult};
 use std::mem;
 use std::ops::{Index, IndexMut};
+use std::sync::atomic::AtomicU16;
 use std::time::SystemTime;
 
 #[cfg(feature = "mock-instant")]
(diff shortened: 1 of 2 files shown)


neptun/src/noise/timers.rs
```diff
@@ -207,6 +230,21 @@ impl Tunn {
...
+            .load(std::sync::atomic::Ordering::Relaxed);
+        for timer_name in TimerName::VALUES {
+            if (timer_mask & (1 << (timer_name as u16))) != 0 {
+                self.timer_tick(timer_name);
+            }
+        }
+        // Reset all marked bits
+        self.timers
+            .timers_to_update_mask
+            .store(0, std::sync::atomic::Ordering::Relaxed);
```
tomaszklak, line 246: I think this will miss `mark_timer_to_update` calls that happen during the for loop. I think you want to call `let timer_mask = self.timers.timers_to_update_mask.swap(0, Relaxed)` in the beginning.

If I'm correct, please also add a test that could catch this issue.

neptun/src/noise/timers.rs
```diff
@@ -65,6 +66,20 @@ pub enum TimerName {
     Top,
 }
 
+impl TimerName {
+    pub const VALUES: [Self; TimerName::Top as usize] = [
+        Self::TimeCurrent,
```
tomaszklak, line 71: Is `TimeCurrent` needed here ?

tomaszklak:
Please add some more detail to the second part:

> This is also needed in order to have a lock free tunnel, because locking timers internally is not possible, otherwise it can lead to a deadlock as update_timers also tries to lock() the timers then.

I don't fully understand this.

neptun/src/noise/timers.rs
Hasan6979: Removed

neptun/src/noise/timers.rs
Hasan6979: Changed and added test

tomaszklak: I don't understand how this test, verifies the fix for the above issue. The issue happens only when the for loop (now in `tick_marked_timers`) runs in parallel with `mark_timer_to_update`. Also, reverting the `swap` change from implementation, doesn't break the test - please add a test that fails with the old version of the code, but passes with the new one.

Hasan6979: Yes, but because of a bit difficulty for running them in parallel to make sure the issue happens, I decided to mark the timer after reading the bit mask. That timer should be updated in the next cycle of update timers, but in previous version that update was being missed.

neptun/src/noise/timers.rs
```diff
@@ -380,3 +417,92 @@ impl Tunn {
...
+        // Mark timers to update
+        my_tun.mark_timer_to_update(super::TimerName::TimeLastDataPacketSent);
+        my_tun.mark_timer_to_update(super::TimerName::TimeLastDataPacketReceived);
+        my_tun.mark_timer_to_update(super::TimerName::TimePersistentKeepalive);
+
+        // Update timers
+        my_tun.update_timers(&mut [0]);
+
+        // Only those timers marked should be udpated
+        assert!(!my_tun.timers[TimerName::TimeLastDataPacketSent].is_zero());
```
packgron, line 450: NIT: use assert_eq!(timers[x], timers[TimeCurrent])

packgron (commented):
+1.0

Hasan6979:
| **ID** | **Role** | **Interval**       | **Transfer** | **Bitrate**    | **Retransmissions** |
|--------|----------|--------------------|--------------|----------------|---------------------|
| 5      | Sender     | 0.00-120.00 sec   | 5.17 GBytes  | 370 Mbits/sec  | 40,779              |
| 5      | Receiver     | 0.00-120.00 sec   | 5.17 GBytes  | 370 Mbits/sec  | N/A                 |
| 7      | Sender     | 0.00-120.00 sec   | 5.21 GBytes  | 373 Mbits/sec  | 42,817              |
| 7      | Receiver     | 0.00-120.00 sec   | 5.21 GBytes  | 373 Mbits/sec  | N/A                 |
Performance test from base

| **ID** | **Role** | **Interval**       | **Transfer** | **Bitrate**    | **Retransmissions** |
|--------|----------|--------------------|--------------|----------------|---------------------|
| 5      | Sender     | 0.00-120.00 sec   | 5.24 GBytes  | 375 Mbits/sec  | 43,217              |
| 5      | Receiver     | 0.00-120.00 sec   | 5.24 GBytes  | 375 Mbits/sec  | N/A                 |
| 7      | Sender     | 0.00-120.00 sec   | 5.08 GBytes  | 364 Mbits/sec  | 41,078              |
| 7      | Receiver     | 0.00-120.00 sec   | 5.08 GBytes  | 364 Mbits/sec  | N/A                 |

jjanowsk:
I've done a ton of testing yesterday on my setup (two physical machines connected directly via cable). Unsurprisingly the most consistent results I get when measuring only one way UDP traffic. On multiple 6 minute runs I got an average of 684 Mbps with this branch vs 662 Mbps with the main branch. Looks like 3% improvement which is nice.

Hasan6979:
@jjanowsk  I think this amount is good enough because I would expect only a marginal gain, if any, by this change. Either way this change is needed for future improvements.

neptun/src/noise/timers.rs
```diff
@@ -82,6 +96,7 @@ pub struct Timers {
     persistent_keepalive: usize,
     /// Should this timer call reset rr function (if not a shared rr instance)
     pub(super) should_reset_rr: bool,
+    timers_to_update_mask: AtomicU16,
```
tomaszklak, line 99: do we need atomicu16 here? can we use instead the `parking_lot::Mutex<u16>`  or even `parking_lot::Mutex<[bool;8]>`?

packgron: So this is main area that should be bringing improment, as then we are just swtiching from a mutex to a mutex.
Now we fundamentaly move away from any wait for lock's.

tomaszklak: This change is adding atomicu16, not removing any mutex

packgron: Oh right, it's not with self as &self, but &mut self.
We don't even need Mutex or Atomic then in reality, just u16 would be enough

jjanowsk:
More updates from the measurements on the CI. When run in the CI on 15 minute runs of single way UDP traffic the gains are not significant. There is less then 0.5% of the difference between main and this branch, sometimes main being faster and sometimes this branch. In total I've run 4 pipelines like this.

tomaszklak:
> Traffic the gains are not significant. There is less then 0.5% of the difference between main and this branch, sometimes main being faster and sometimes this branch.

In that case, maybe we can close this PR?

PR #4, opened 2024-11-29
Add initial xray implementation

mathiaspeters:
`README.md` should have all info needed to review


xray/README.md
```diff
@@ -0,0 +1,53 @@
+# X-Ray
+
+A tool to inspect the performance of different wireguard implementations
+
+## How it works
+
+The setup consists of two clients and a wireguard interface. One of the clients sends and receives encrypted packets (let's call this the crypto socket), and the other sends and receives plaintext packets (let's call this the plaintext socket). The packets that are being sent contain a sequence number so that the receivning socket can now which packet it is, and a timestamp of when it was sent.
```
lcruz99, line 7: nit: recevning -> receiving

xray/run.py
```diff
@@ -0,0 +1,145 @@
...
+
+    wg = Wireguard.from_str(args.wg)
+    test_type = args.test_type.lower() if args.test_type is not None else "crypto"
+    assert test_type in [
+        "crypto",
+        "pt",
+    ], f"Invalid test type '{test_type}'. Valid options are 'crypto' and 'pt'"
+    count = int(args.count) if args.count is not None else 10
+    assert count > 0, f"Count must be at least one, but got {count}"
+    build_neptun = args.nobuild is None
```
lcruz99, line 114: You're comparing boolean with None, will always result False.

xray/run.py
```diff
@@ -0,0 +1,145 @@
...
+    )
+
+
+def stop_tcpdump(tcpdump):
+    run_command(f"kill {tcpdump.pid}")
+
+
+def destroy_wireguard(wg):
+    if wg == Wireguard.NepTUN:
+        run_command("killall neptun-cli")
```
lcruz99, line 91: While testing, I ended up with the `neptun-cli` processes hanging, and manually running this (SIGTERM) didn't kill those hanging processes, only SIGKILL worked. So I'd suggest running `killall -9 neptun-cli` here instead.

lcruz99: same below for the Wireguard.BoringTun` case

xray/run.py
```diff
@@ -0,0 +1,145 @@
...
+    except:  # noqa: E722
+        pass
+
+    setup_wireguard(wg, build_neptun)
+    tcpdump = start_tcpdump(get_pcap_name(wg.name, test_type, count))
+
+    succeeded = True
+    try:
+        run_xray(wg.name, test_type, count)
+    except:  # noqa: E722
```
lcruz99, line 129: when does cargo run thrown an exception?

In case the test goes wrong, let's say the receiver doesn't get any packets, cargo process will still succeed which later will make `analyze()` fail.

I can't think right away of an alternative to check if we should proceed with analyze() but probably there is

mathiaspeters: It was a simple as adding `check=True` to the `subprocess.run` command in my `run_command` function

xray/run.py
mathiaspeters: I think it would be `None` if the argument wasn't passed. Good catch!

xray/src/main.rs
```diff
@@ -0,0 +1,325 @@
...
+        &mut self,
+        cmd: TestCmd,
+        finish_timeout: &mut std::pin::Pin<&mut tokio::time::Sleep>,
+    ) {
+        match cmd {
+            TestCmd::Done => {
+                println!("All packets were sent. Waiting for maximum 10 seconds to receive");
+                finish_timeout
+                    .as_mut()
+                    .reset(Instant::now() + Duration::from_secs(10));
```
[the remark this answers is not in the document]
mathiaspeters: I did that at first but I was having problems with tcpdump not gathering all packets. 10s is most likely overkill in this scenario so I could reduce it to 3-4 seconds after the last packet has been received, or 10s if not all packets have been received

xray/README.md
```diff
@@ -0,0 +1,55 @@
...
+## Running it
+
+X-Ray currently only works on linux
+
+Nordlayer and/or nordvpn might interfere with xray, so try disabling them if you're not receving any of the packets
+
+### Support
+
+Supported adapter types:
+- Naitve
```
lcruz99, line 30: Naitve -> Native

xray/src/main.rs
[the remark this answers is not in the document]
lcruz99: fair enough 👍

xray/README.md
```diff
@@ -0,0 +1,55 @@
...
+
+The application is executed with the `run.py` script. I takes some arguments, all of which are optional:
+
+- **--wg <adapter type>**: which adapter type to use, accepts one of `neptun`, `native`, `wggo` and `boringtun`. Default is `neptun`
+
+- **--test_type <test type>**: which kind of test to run, accepts `crypto` (to send packets from the crypto socket to the plaintext socket) and `pt` (to send packets from the plaintext socket to the crypto socket). Default is `crypto`
+
+- **--count <number of packets>**: How many packets to send. Default is 10
+
+- **--nobuild**: whether or not to build `neptun-cli` before running. Only relevant when the adaptert type is `neptun`. Default is to build `neptun`
```
lcruz99, line 45: Now outdated with the new `--nobuild-neptun` and `--nobuild-xray`

xray/src/client.rs
```diff
@@ -0,0 +1,222 @@
...
+            ipv4.set_destination(*dst.ip());
+            ipv4.set_checksum(ipv4::checksum(&ipv4.to_immutable()));
+        }
+        {
+            let mut udp = MutableUdpPacket::new(&mut udp_packet[20..]).unwrap();
+            udp.set_source(src.port());
+            udp.set_destination(dst.port());
+            udp.set_length(8 + payload.len() as u16);
+            udp.set_payload(payload);
+            // TODO: Calculate checksum
```
lcruz99, line 205: Given the new review guidelines, this TODO should contain your name or the ticket number:
// TODO(mathias):  or // TODO(LLT-9876):

xray/README.md
```diff
@@ -0,0 +1,55 @@
+# X-Ray
+
+A tool to inspect the performance of different wireguard implementations
+
+## How it works
+
+The setup consists of two clients and a wireguard interface. One of the clients sends and receives encrypted packets (let's call this the crypto socket), and the other sends and receives plaintext packets (let's call this the plaintext socket). The packets that are being sent contain a sequence number so that the receiving socket can now which packet it is, and a timestamp of when it was sent.
```
packgron, line 7:
```suggestion
The setup consists of two sockets and a wireguard interface. One of the socket sends and receives encrypted packets (let's call this the crypto socket), and the other sends and receives plaintext packets (let's call this the plaintext socket). The packets that are being sent contain a sequence number so that the receiving socket can now which packet it is, and a timestamp of when it was sent. 
```

xray/README.md
```diff
@@ -0,0 +1,55 @@
...
+- Naitve
+- Neptun
+- WireguardGo (needs to be installed beforehand)
+- Boringtun (needs to be compiled and placed in the `neptun/target/release` folder beforehand)
+
+### Running the test
+
+The application is executed with the `run.py` script. I takes some arguments, all of which are optional:
+
+- **--wg <adapter type>**: which adapter type to use, accepts one of `neptun`, `native`, `wggo` and `boringtun`. Default is `neptun`
```
packgron, line 39: Maybe it makes sense to also allow use of a bin path? to xplatform interface compliant  implementation of wg

packgron (commented):
Initial review of rust part,
I fell that at some point it would actualy be a good idea to have bidirectional testing.

As we know that rx and tx cannot happen simuntaniosly in neptun, maybe it causes issues

xray/README.md
mathiaspeters: I think that's generally a good idea though I think there's more work that will need to be done to actually make xray cross platform. I would do this as a separate PR

xray/src/utils.rs
```diff
@@ -0,0 +1,167 @@
...
+
+pub async fn configure_wg(
+    adapter_type: &str,
+    wg_name: &str,
+    wg_keys: &KeyPair,
+    peer_keys: &KeyPair,
+    wg_port: u16,
+    ips: &[Ipv4Addr],
+) {
+    for ip in ips {
```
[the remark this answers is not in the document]
mathiaspeters: I'm currently not setting the MTU, because the packets are nowhere near 1420 bytes, but [here](https://github.com/NordSecurity/NepTUN/pull/4/files#diff-26b7434fdea764dccd8147b3e910f915c3e275970ec3faeab8f373260a7ea222R64) `ip link up` is called

xray/README.md
mathiaspeters: I created LLT-5845 for it

xray/src/main.rs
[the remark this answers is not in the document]
mathiaspeters: Fixed

xray/README.md
mathiaspeters: Fixed

xray/README.md
```diff
@@ -0,0 +1,55 @@
...
+The setup consists of two clients and a wireguard interface. One of the clients sends and receives encrypted packets (let's call this the crypto socket), and the other sends and receives plaintext packets (let's call this the plaintext socket). The packets that are being sent contain a sequence number so that the receiving socket can now which packet it is, and a timestamp of when it was sent. 
+
+Execution consists of these steps:
+1. The wireguard interface is started
+2. `tcpdump` is started, with packets being written to a pcap file
+3. The main X-Ray executable is started
+  3.1. The crypto client performs a handshake with the wireguard interface
+  3.2. The specified number of packets are sent in the specified direction (crypto->wg interface->plaintext, or plaintext->wg interface->crypto)
+  3.3. The receiving socket registers the packets it receives that are part of the test
+  3.4. The packets are written to a `.csv` file with each row containing the sequence number, the time it was sent and, if it was received, the time it was received
```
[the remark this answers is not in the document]
mathiaspeters: Fixed

xray/README.md
mathiaspeters: Fixed

xray/src/client.rs
mathiaspeters: Fixed

xray/src/utils.rs
```diff
@@ -0,0 +1,167 @@
...
+}
+
+// pub async fn read_from_uapi(sock: &UnixStream) -> String {
+//     sock.readable().await.unwrap();
+//     let mut buf = vec![0; 1024];
+//     let bytes_read = sock.try_read(&mut buf).unwrap();
+//     buf.truncate(bytes_read);
+
+//     String::from_utf8(buf).unwrap()
+// }
```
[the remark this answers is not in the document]
mathiaspeters: Sort of. I kept it there but commented out in case it would be needed in the future, but it makes more sense to remove it for now, so I did that

xray/README.md
mathiaspeters: Fixed

xray/src/key_pair.rs
```diff
@@ -0,0 +1,54 @@
...
+use neptun::x25519::{PublicKey, StaticSecret};
+
+pub struct KeyPair {
+    pub private: StaticSecret,
+    pub public: PublicKey,
+}
+
+impl KeyPair {
+    pub fn new() -> Self {
+        let private = StaticSecret::random_from_rng(rand::thread_rng());
```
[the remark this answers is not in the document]
mathiaspeters: Fixed

xray/src/client.rs
```diff
@@ -0,0 +1,222 @@
...
+            tunn,
+            sock,
+            buf: vec![0; 4096],
+        }
+    }
+
+    pub async fn do_handshake(&mut self, wg_addr: SocketAddrV4) {
+        println!("Handshake: starting");
+        let tunn = self.tunn.as_mut().unwrap();
+        match tunn.format_handshake_initiation(&mut self.buf, true) {
```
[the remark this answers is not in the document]
mathiaspeters: Yes. When just relying on the handshake being done as part of the first test packet it was taking a long time to do the handshake and it required special handling of stopping the flow of packets until the handshake was done so it wouldn't lose all but 256 packets. It was possible, but this felt much cleaner

xray/src/main.rs
```diff
@@ -0,0 +1,315 @@
...
+        &peer_keys,
+        wg_addr.port(),
+        &[*wg_addr.ip()],
+    )
+    .await;
+
+    let (cmd_tx, cmd_rx) = mpsc::channel::<TestCmd>(100);
+    let (done_tx, mut done_rx) = mpsc::channel::<()>(1);
+
+    let pt_sock = UdpSocket::bind(pt_addr).await.unwrap();
```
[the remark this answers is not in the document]
mathiaspeters: Renamed to `plaintext_sock`

xray/src/main.rs
```diff
@@ -0,0 +1,315 @@
...
+    let pt_client = Client::new(pt_addr, None, pt_sock);
+
+    let crypto_sock = UdpSocket::bind(crypto_sock_addr).await.unwrap();
+    let tunn = Tunn::new(peer_keys.private, wg_keys.public, None, None, 123, None).unwrap();
+    let mut crypto_client = Client::new(crypto_addr, Some(tunn), crypto_sock);
+    crypto_client.do_handshake(wg_addr).await;
+
+    let mut event_loop =
+        EventLoop::new(cli_args, wg_addr, crypto_client, pt_client, done_tx, cmd_rx);
+    let task = tokio::task::spawn(async move {
```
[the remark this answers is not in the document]
mathiaspeters: Fixed

xray/src/main.rs
```diff
@@ -0,0 +1,315 @@
...
+        wg_name,
+        &wg_keys,
+        &peer_keys,
+        wg_addr.port(),
+        &[*wg_addr.ip()],
+    )
+    .await;
+
+    let (cmd_tx, cmd_rx) = mpsc::channel::<TestCmd>(100);
+    let (done_tx, mut done_rx) = mpsc::channel::<()>(1);
```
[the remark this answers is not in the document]
mathiaspeters: I tried just awaiting the join handle before and it was exiting way too early which is why I added that channel in the first place, but I must have made some significant changes because I removed the channel now and it still works

xray/src/main.rs
```diff
@@ -0,0 +1,315 @@
...
+        }
+    }
+    cmd_tx.send(TestCmd::Done).await.unwrap();
+
+    done_rx.recv().await.unwrap();
+
+    task.await.unwrap();
+}
+
+struct EventLoop {
```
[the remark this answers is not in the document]
mathiaspeters: Yup, moved to its own file

xray/src/main.rs
```diff
@@ -0,0 +1,315 @@
...
+            is_done: false,
+            crypto_buf: vec![0; 1024],
+            pt_buf: vec![0; 1024],
+            recv_packet: 1,
+        }
+    }
+
+    async fn run(&mut self) {
+        let mut wg_tick_interval = tokio::time::interval(Duration::from_millis(250));
+        let finish_timeout = tokio::time::sleep(Duration::from_secs(1));
```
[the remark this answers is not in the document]
mathiaspeters: That timeout is only used when the final packet has been sent, and at that point the timeout is set to something more reasonable. The 1 second here is just to initialize it to something.
I added a comment in the code to clarify it

xray/src/main.rs
```diff
@@ -0,0 +1,315 @@
...
+    async fn on_recv_plaintext_packet(&mut self, rt: RecvType) -> bool {
+        if let RecvType::Data {
+            length: bytes_read, ..
+        } = rt
+        {
+            if bytes_read == Packet::send_size() {
+                if self.recv_packet % (self.cli_args.packet_count / 10) == 0 {
+                    println!("[Plaintext] Received {} packets", self.recv_packet);
+                }
+                let index = u64::from_le_bytes(self.pt_buf[0..8].try_into().unwrap());
```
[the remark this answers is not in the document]
mathiaspeters: All the timestamps are absolute epoch timestamps, so the receiver doesn't need to know what the send time was. Both timestamps are written to the csv file and analyzed separately

mathiaspeters:
> Looking at the plot we get 3 top right empty subgraphs: 
> 
> Is this intended?

Yes and no. The intention was to have one column of the graph for csv analysis and one for pcap analysis, so the intention was that they might have different number of graphs, but matplotlib apparently populates it with empty graphs so both sides have the same number of graphs. I changed it so it will distribute all defined graphs evenly. There might still be one empty graph going forward if the number of defined graphs is odd

mathiaspeters:
> I fell that at some point it would actualy be a good idea to have bidirectional testing.

I was thinking about that before too but didn't implement it. I created LLT-5844 for it

xray/src/main.rs
```diff
@@ -0,0 +1,315 @@
...
+    utils::{configure_wg, write_to_csv, Packet, RecvType, SendType, TestCmd},
+};
+
+struct CliArgs {
+    packet_count: usize,
+    csv_name: String,
+}
+
+#[tokio::main]
+async fn main() {
```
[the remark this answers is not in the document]
mathiaspeters: Looks interesting, I'll see how it looks for xray

xray/src/client.rs
mathiaspeters: I just implemented the checksum so the `TODO` was removed

xray/src/main.rs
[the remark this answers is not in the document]
mathiaspeters: It seemed like the amount of work to use `color-eyre` would be similar to adding standard error handling so I went with a more standard solution

xray/src/client.rs
```diff
@@ -0,0 +1,231 @@
...
+        let packet = Self::make_udp_packet(self.addr, packet_dst, payload)?;
+        let tr = tunn.encapsulate(&packet, &mut self.buf);
+        let (send_type, packet): (SendType, &[u8]) = match tr {
+            TunnResult::Done => (SendType::None, &[]),
+            TunnResult::WriteToNetwork(p) => match p[0] {
+                1 => (SendType::Handshake, p),
+                4 => (SendType::Data, p),
+                unexpected => return Err(XRayError::UnexpectedPacketType(unexpected)),
+            },
+            TunnResult::WriteToTunnelV4(p, _) => (SendType::Tunn, p),
```
packgron, line 126: We should not recieve this (basically it' unexpected), and also, if we do, it should not go into crypto socket, as it's a plaintext packet, intended fot the tunnel

xray/src/client.rs
```diff
@@ -0,0 +1,231 @@
...
+    }
+
+    async fn recv_encrypted(&mut self, buf: &mut [u8]) -> XRayResult<RecvType> {
+        let (mut bytes_read, from) = self.sock.recv_from(&mut self.buf).await?;
+        let from = match from {
+            SocketAddr::V4(addr) => addr,
+            SocketAddr::V6(_) => return Err(XRayError::Ipv6),
+        };
+        let mut ret = match self.buf[0] {
+            2 => RecvType::Handshake,
```
packgron, line 147: So you always expect this side to be initiator and then a receiver? It look a bit confusing to have 1 and 2 map to handshake instead of HandshakeInitiation / HandshakeResponse

xray/src/event_loop.rs
```diff
@@ -0,0 +1,213 @@
...
+                .duration_since(std::time::UNIX_EPOCH)?
+                .as_micros();
+            let mut packet = Packet {
+                index,
+                send_ts,
+                recv_ts: None,
+            };
+            packet.serialize(&mut payload, 0);
+
+            packet.index = 0;
```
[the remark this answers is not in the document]
packgron: Oh i guess it is identified by position, and the you can check if postion matches index ?

xray/src/main.rs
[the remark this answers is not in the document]
packgron: So color_eyre, can work with thiserror, it just gives a "nicer" report, and also backtrace :D it's also just a for of anyhow, but i started to like this one a bit more.

xray/src/utils.rs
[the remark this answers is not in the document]
mathiaspeters: It's now set in `run.py`

xray/src/client.rs
mathiaspeters: Good point. That case is now considered `unexpected`

xray/src/client.rs
mathiaspeters: The idea was to be the initiator of handshakes but it should of course work with being the receiver as well. I changed the `SendType` and `RecvType` to have `HandshakeInitiation` and `HandshakeResponse` instead of just `Handshake`. I haven't actively tried if it works to be on the responder side of a handshake but I think it should work

xray/src/event_loop.rs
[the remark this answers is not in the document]
mathiaspeters: I renamed the different places where the index is used to `send_index` and `recv_index`,  and updated this part of the code to be a bit clearer around the index handling

packgron (approved):
+1.0

packgron (approved):

PR #5, opened 2024-12-02
LLT-5809

jjanowsk:
Some cleanup of UAPI handler. Details in commit messages.

### DIFF
--- a/neptun/src/device/api.rs
+++ b/neptun/src/device/api.rs
@@ -230,18 +230,21 @@ fn api_set<R: Read>(reader: &mut BufReader<R>, d: &mut LockReadGuard<Device>) ->
                             },
                             Err(_) => return EINVAL,
                         },
-                        #[cfg(any(
-                            target_os = "android",
-                            target_os = "fuchsia",
-                            target_os = "linux"
-                        ))]
-                        "fwmark" => match val.parse::<u32>() {
-                            Ok(mark) => match device.set_fwmark(mark) {
-                                Ok(()) => {}
-                                Err(_) => return EADDRINUSE,
-                            },
-                            Err(_) => return EINVAL,
-                        },
+                        "fwmark" =>
+                        {
+                            #[cfg(any(
+                                target_os = "android",
+                                target_os = "fuchsia",
+                                target_os = "linux"
+                            ))]
+                            match val.parse::<u32>() {
+                                Ok(mark) => match device.set_fwmark(mark) {
+                                    Ok(()) => {}
+                                    Err(_) => return EADDRINUSE,
+                                },
+                                Err(_) => return EINVAL,
+                            }
+                        }
                         "replace_peers" => match val.parse::<bool>() {
                             Ok(true) => device.clear_peers(),
                             Ok(false) => {}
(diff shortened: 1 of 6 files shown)


neptun/src/device/integration_tests/mod.rs
```diff
@@ -477,6 +473,33 @@ mod tests {
         assert!(response.ends_with("errno=0\n\n"));
     }
 
+    #[test]
+    #[ignore]
```
tomasz-grz, line 477: Is it right that we don't want to normally run these tests?

neptun/src/device/integration_tests/mod.rs
```diff
@@ -477,6 +473,33 @@ mod tests {
...
+        let response = wg.wg_uapi_cmd("get=1\n\nget=1\n\n");
+        assert_eq!(response.matches("errno=0\n\n").count(), 2);
+
+        let response = wg.wg_uapi_cmd("get=1\n\nset=1\nlisten_port=12345\n\nget=1\n\n");
+        assert_eq!(response.matches("errno=0\n\n").count(), 3);
+    }
+
+    #[test]
+    #[ignore]
+    fn test_wireguard_uapi_set_without_proper_ending() {
```
tomasz-grz, line 490: What do you think about some tests that respond with errors?

neptun/src/device/api.rs
```diff
@@ -148,26 +114,31 @@ pub fn api_exec<R: Read, W: Write>(
...
-        cmd.pop(); // pop the new line character
-        let status = match d.closed {
+    while status == 0 && reader.read_line(&mut cmd).is_ok_and(|n| n > 0) {
+        status = match d.closed {
             true => ENOENT,
             false => match cmd.as_ref() {
                 // Only two commands are legal according to the protocol, get=1 and set=1.
-                "get=1" => api_get(writer, d),
-                "set=1" => api_set(reader, d),
+                "get=1\n" => api_get(reader, writer, d),
```
tomasz-grz, line 124: Do we really want to deal with all these new lines manually? What do you think about using something like `cmd.trim()` to sanitize the input, and `cmd.split_whitespace()` if there are multiple commands?

tomasz-grz (commented):
Nice work, some questions since I'm not too familiar with the protocol.

dfetti (commented):
+1.0 I'm not familiar with the project, but looks good to me. 
Might be a good idea to start a changelog file for this project?

neptun/src/device/integration_tests/mod.rs
jjanowsk: That's just the way Boringtun used to differentiate integration tests. They are marked as ignored and you run them via `cargo tests --ignored`

neptun/src/device/api.rs
jjanowsk: Yes, we do. It's not about striping them. Stripping was (incorrectly) done earlier via `pop`. Here the point is that the \n is required. So sending "get=1[EOF]" is considered an incorrect command. It's just to keep compatibility with wireguard-go, nothing more.

neptun/src/device/integration_tests/mod.rs
jjanowsk: Could add a few, but TBH I don't think there would be much value out of them.

tomasz-grz (approved):
+1 nice work

PR #6, opened 2024-12-05
LLT-5652 Performance measurements

jjanowsk:
First version of performance measurement rig

Runs iperf3 via wireguard-go, main branch of NepTUN and the current
branch between two docker containers. The results are printed in the
job log. Probably we want to add some configurability to those scripts
but it is better to do it as we go, when we actually know what we need.

Most likely we want those improvements:
- host docker image and make automation for its update

### DIFF
--- a/xtask/perf/left.sh
+++ b/xtask/perf/left.sh
@@ -1,0 +1,47 @@
+#!/bin/bash
+
+wireguard-go wg0
+wg set wg0 \
+    listen-port 51820 \
+    private-key <(echo AH7jKt6M0RS21MkRG+URrfgmwvJqdhVvqtKeF4WR+E8=) \
+    peer mqrc8+LD+6zvMeCtyCcIBPEYuXT74lq1Hros0Np8ZgA= \
+    allowed-ips 10.0.0.2/32 \
+    endpoint 176.0.0.3:51820
+ip address add dev wg0 10.0.0.1/24
+ip link set up dev wg0
+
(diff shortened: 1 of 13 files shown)


gytsto (commented):
looks great +1

Hasan6979:
Can you add tcp tests as well ?

xtask/perf/left.sh
```diff
@@ -0,0 +1,47 @@
...
+    private-key <(echo 0Fn5JWI1QGDiaVYLDBSLklIEBUujfpX1oH/UGI2D62k=) \
+    peer fqrU1Wk8nuyy6phf8IUDwcKK1ElslFYDyteANr2hlgM= \
+    allowed-ips 10.0.2.2/32 \
+    endpoint 176.0.0.3:51822
+ip address add dev wg2 10.0.2.1/24
+ip link set up dev wg2
+
+echo
+echo "Raw network:"
+iperf3 -i 10 -t  10 --bidir -c 176.0.0.3
```
Hasan6979, line 35: Add `-b 0` to really try to throttle the connection

jjanowsk: This is actually TCP, so `-b` is not needed. Regarding adding second protocol I think we can add it later, given that we want to merge this one rather sooner then later

Hasan6979: Ah. yes. right!

Hasan6979:
+1

Hasan6979 (approved):

PR #10, opened 2024-12-13
Fix DeviceHandle::send_uapi_cmd

jjanowsk:
Previous implementation of api_exec was not flushing the writter after writing to it. The new one does and hance the internal buffer of BufWriter is cleared. That caused send_uapi_cmd to always return empty string. With this fix we correctly return the external buffer to which BufWritter is writing.


packgron (commented):
+1.0

ThrasherLT (approved):
Tested it out again and the tests pass, so +1.0

PR #11, opened 2024-12-13
XRay bug fixes

mathiaspeters:


xray/README.md
```diff
@@ -18,6 +18,12 @@ Execution consists of these steps:
 5. The wireguard interface is destroyed
 6. The `.csv` file and pcap are analyzed
 
+The structure and exeuction of the XRay binary can be represented with the following image:
```
mathiaspeters, line 21: Adding diagram from presentation to the readme

xray/README.md
```diff
@@ -48,8 +54,6 @@ The application is executed with the `run.py` script. I takes some arguments, al
 
 ## Known issues
 
--  The analysis of pcaps is quite limited right now because it doesn't decrypt the packets (this is being worked on)
+- The analysis of pcaps is quite limited right now because it doesn't decrypt the packets (this is being worked on)
 
 - There are multiple inefficiencies that could potentially impact the test results, the main one being not reusing buffers when creating and sending packets. Each packet that gets constructed allocates a new buffer when they could all reuse the same one
-
```
mathiaspeters, line 54: Should not be relevant anymore with retry logic

xray/analyze.py
```diff
@@ -22,7 +22,8 @@ def __init__(self, csv_path):
             reader = csv.reader(csvfile, delimiter=",", quotechar="|")
             next(reader)
             for row in reader:
-                self.indices.append(int(row[0]))
+                if len(row[0]) > 0:
```
mathiaspeters, line 25: When the previous code was added a receive index of 0 meant that the packet hadn't arrived, but it's better represented with empty data. After making that change, this part of the code wasn't updated to handle empty index fields which would cause exceptions when some packet(s) didn't arrive

xray/analyze.py
```diff
@@ -147,7 +148,12 @@ def packet_latency(self, ax):
         ax.hist(buckets, color="blue", bins=num_buckets)
 
     def dropped_packets(self, ax):
-        num_buckets = 100
+        if self.count >= 100:
```
mathiaspeters, line 151: If the number of packets (`self.count`) was less than 100, the `bucket_size` would be 0, and the subsequent `/ bucket_size` on line 161 would lead to a division by zero exception. This adjusts the bucket sizes to work for smaller sample sizes

xray/src/event_loop.rs
```diff
@@ -42,7 +42,7 @@ impl EventLoop {
             is_done: false,
             crypto_buf: vec![0; 1024],
             plaintext_buf: vec![0; 1024],
-            recv_counter: 1,
+            recv_counter: 0,
```
mathiaspeters, line 45: This makes more sense with the debug prints, or rather, doing it this way allows for better debug prints without having to do a `-1` whenever it is referenced. A `+1` is needed in a couple of places instead, but significantly fewer than the number of `-1`s that would be otherwise needed

xray/src/event_loop.rs
```diff
@@ -55,7 +55,7 @@ impl EventLoop {
         loop {
             tokio::select! {
                 _ = &mut finish_timeout, if self.is_done => {
-                    self.on_finished(self.packets.len()).await?;
+                    self.on_finished(self.recv_counter).await?;
```
mathiaspeters, line 58: This was printing that all packets were received even if none were, since `self.packets` contains all sent packets instead of all received packets (receiving a packet will just update the packet in the vector since it's basically storing metrics)

xray/src/event_loop.rs
```diff
@@ -190,19 +191,18 @@ impl EventLoop {
...
 
     fn on_maybe_recv_all(&self, finish_timeout: &mut Pin<&mut tokio::time::Sleep>) {
-        if self.recv_counter > self.cli_args.packet_count {
-            println!("All packets were received. Waiting 3 seconds to make sure pcap is properly populated");
+        if self.recv_counter >= self.cli_args.packet_count {
+            println!("All packets were received. Finishing test...");
             finish_timeout
                 .as_mut()
-                .reset(Instant::now() + Duration::from_secs(3));
+                .reset(Instant::now() + Duration::from_secs(1));
```
mathiaspeters, line 205: This might not be necessary at all going forward, at least it's not necessary at the moment, but it might be necessary in a separate PR so reducing the timeout is better for now than removing the timeout completely

xray/src/client.rs
```diff
@@ -33,34 +33,39 @@ impl Client {
...
+                tokio::select! {
+                    Ok(recv_type) = self.recv_encrypted(&mut handshake_buf) => {
+                        if matches!(recv_type, RecvType::HandshakeResponse) {
+                            println!("Handshake: done");
+                            return Ok(());
+                        }
+                    }
+                    _ = &mut handshake_timeout => {
+                        println!("Handshake timed out trying again...");
+                        continue 'new_handshake;
```
tomaszklak, line 62: I know this was already here, but - can this be simplified and use https://docs.rs/tokio/latest/tokio/time/fn.timeout.html ?

mathiaspeters: Fixed

xray/src/client.rs
```diff
@@ -33,34 +33,39 @@ impl Client {
...
-            .tunn
-            .as_mut()
-            .expect("This function should only be called on clients with a Tunn object");
-        match tunn.format_handshake_initiation(&mut self.buf, true) {
-            TunnResult::WriteToNetwork(packet) => {
-                self.sock.send_to(packet, wg_addr).await?;
-            }
-            unexpected => {
-                return Err(XRayError::UnexpectedTunnResult(format!("{unexpected:?}")));
+        'new_handshake: for _ in 0..3 {
```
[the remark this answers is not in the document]
mathiaspeters: I think that `futures-retry` is a bit overkill here but I like the suggestion to split the code here, and I introduced some constants to make it a bit cleaner. Let me know what you think

mathiaspeters:
> Can your remove the png and replace it with mermaid chart?

Fixed. Please have a look

matislovas:
+1

tomaszklak (approved):
+1

PR #14, opened 2024-12-18
Add pcap analysis to XRay

mathiaspeters:
This PR adds pcap analysis to XRay, though it's likely still not utilized as much as it could, but the infrastructure is now in place.

How it works is that the pcap file is opened in rust when the test is done, since we have the `noise::Tunn` object there that encrypted the packets, and we decrypt the packets after the fact. This requires adding some functions to noise itself, but those are feature gated so they can't be used in the wrong place. The data that is gathered from the pcaps (currently the timestamp provided by tcpdump and the sender index) are written to the CSV file, and then used in the analysis step. The previous histograms have been replaced by stacked historgrams so that the different checkin points can be seen separately.

### DIFF
--- a/neptun/src/noise/mod.rs
+++ b/neptun/src/noise/mod.rs
@@ -331,6 +331,43 @@ impl Tunn {
         self.handle_verified_packet(packet, dst)
     }
 
+    #[cfg(feature = "xray")]
+    pub fn decrypt<'a>(
+        &mut self,
+        datagram: &[u8],
+        dst: &'a mut [u8],
+    ) -> Result<&'a [u8], WireGuardError> {
+        let packet = Tunn::parse_incoming_packet(datagram)?;
+        match packet {
+            Packet::PacketData(p) => {
+                let r_idx = p.receiver_idx as u32;
+
+                // Get the (probably) right session
+                let decapsulated_packet = {
+                    let session = self.sessions.iter().find_map(|s| match s {
+                        Some(session) if session.is_right_session(r_idx) => Some(session),
+                        _ => None,
+                    });
+                    let session = session.ok_or_else(|| {
+                        tracing::trace!(
+                            message = "No current session available",
+                            remote_idx = r_idx
+                        );
+                        WireGuardError::NoCurrentSession
+                    })?;
+                    session.decrypt_data_packet(p, dst)?
+                };
+
+                match self.validate_decapsulated_packet(decapsulated_packet) {
+                    TunnResult::WriteToTunnelV4(p, _) => Ok(p),
+                    TunnResult::Err(err) => Err(err),
+                    _ => Err(WireGuardError::UnexpectedPacket),
+                }
+            }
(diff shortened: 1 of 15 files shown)


xray/src/pcap.rs
```diff
@@ -0,0 +1,74 @@
...
+pub fn process_pcap(pcap_path: &str, mut tunn: Tunn) -> XRayResult<Vec<PcapPacket>> {
+    let mut packets = Vec::new();
+
+    let mut capture = Capture::from_file(pcap_path)?;
+    let mut decrypt_buf = vec![0; 1024];
+
+    'top_level: while let Ok(packet) = capture.next_packet() {
+        let ts = packet.header.ts;
+        let ts = ts.tv_sec as u128 * 1_000_000 + ts.tv_usec as u128;
+        if let Some(sll2_packet) = SLL2Packet::new(packet.data) {
```
packgron, line 27: I think switching to early returns will minify the "Indent Hadouken"

packgron: Well more like early continues, or you could also extract singular packet parsing into a seperate function

xray/src/pcap.rs
```diff
@@ -0,0 +1,74 @@
...
+    pub send_index: u64,
+}
+
+pub fn process_pcap(pcap_path: &str, mut tunn: Tunn) -> XRayResult<Vec<PcapPacket>> {
+    let mut packets = Vec::new();
+
+    let mut capture = Capture::from_file(pcap_path)?;
+    let mut decrypt_buf = vec![0; 1024];
+
+    'top_level: while let Ok(packet) = capture.next_packet() {
```
packgron, line 24: label is mostly for clarity?, as i don't see any other iteration

neptun/src/noise/mod.rs
```diff
@@ -331,6 +331,41 @@ impl Tunn {
         self.handle_verified_packet(packet, dst)
     }
 
+    #[cfg(feature = "xray")]
+    pub fn decrypt<'a>(
```
packgron, line 335: So you can only use this while tunnel is "hot" right ?
Basicaly we have a 180s * 3 (you can see in timers)
I guess this is fine for now, but would be nice if we could rebuild everything from packet's in long past :D, let's at minimum have a jira ticket to track it

mathiaspeters: I had this thought too but didn't really want to dig into it, but it's a good idea :smile: Created LLT-5881 for it

xray/src/pcap.rs
mathiaspeters: You don't like 7-8 levels of indentation?! :smile: I extracted it into a few function with early returns where practical

xray/src/pcap.rs
mathiaspeters: Leftover from a previous version that had nested loops. Removed

packgron (commented):
+1.0

gytsto (approved):
+1

PR #15, opened 2024-12-18
Fix Windows CI

jjanowsk:
For some reason cargo-hack does not have a Cargo.lock file and currently one of it's dependencies requires newer version of rust that we use. Let's use latest stable for now.


Hasan6979:
+2

Hasan6979 (approved):

mathiaspeters (approved):
+2

PR #16, opened 2024-12-18
LLT-5878: Xray tests on CI

tomasz-grz:
This MR adds a `xray-tests` github workflow that performs the tests and displays the result in text format, as well as uploading the `txt` or `png` graphs as artifacts. 

To achieve this 3 more arguments were added to xray run script:
- `--save-output` saves the analysis charts in `results` folder to be uploaded as job artifacts.
- `--ascii` output the analysis chart as text graph to be displayed in the CI job.
- `--disable-drop-privileges` flag is passed to `neptun-cli` and `boringtun-cli`, without it NepTUN is failing to start on the CI runner.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -49,3 +49,32 @@ jobs:
         fetch-depth: 0
     - run: docker build -t neptun-runner:0.0.1 .
     - run: cargo xtask perf --base main
+
+  xray-tests:
+    runs-on: ubuntu-24.04
+    steps:
+      - name: Checkout
+        uses: actions/checkout@a5ac7e51b41094c92402da3b24376905380afc29
+      - name: Setup
+        working-directory: xray
+        run: |
+          sudo apt update
+          sudo apt install -y wireguard wireguard-go
+          python -m pip install --upgrade pip
+          pip install pipenv
+          pipenv install
+      - name: Execute xray
+        working-directory: xray
+        run: |
+          pipenv run python run.py --wg native --ascii --save-output --test-type crypto --count 10000
+          pipenv run python run.py --wg native --ascii --save-output --test-type plaintext --count 10000
+          pipenv run python run.py --wg neptun --ascii --save-output --test-type crypto --count 10000 --disable-drop-privileges
+          pipenv run python run.py --wg neptun --ascii --save-output --test-type plaintext --count 10000 --disable-drop-privileges
+      - name: Results
+        working-directory: xray/results
+        run: |
+          for file in *.txt; do
+            echo "----- $file -----"
+            cat "$file"
+            echo ""
+          done
--- a/xray/README.md
+++ b/xray/README.md
@@ -69,6 +69,12 @@ The application is executed with the `run.py` script. I takes some arguments, al
 
 - **--nobuild-xray**: whether or not to build `xray` before running. Default is to build `xray`
 
+- **--save-output**: save the analysis charts in `results` folder.
+
+- **--ascii**: output the analysis chart as text graph.
+
+- **--disable-drop-privileges**: pass the `disable-drop-privileges` flag to `neptun-cli` and `boringtun-cli`.
+
 ## Known issues
 
 - The analysis of pcaps is quite limited right now because it doesn't decrypt the packets (this is being worked on)
--- a/xray/analyze.py
+++ b/xray/analyze.py
@@ -1,13 +1,17 @@
 import csv
 import math
-import matplotlib.pyplot as plt  # type: ignore
 from functools import reduce
+
+import matplotlib as mpl
+import matplotlib.pyplot as plt  # type: ignore
+import mpl_ascii
+from paths import PathGenerator
 from scapy.all import PcapReader  # type: ignore
 from scapy.layers.inet import UDP  # type: ignore
 
 
-def analyze(csv_path, pcap_path, count, test_type):
-    Analyzer(csv_path, pcap_path, count, test_type)
+def analyze(paths, count, test_type, ascii, save_output):
+    Analyzer(paths, count, test_type, ascii, save_output)
 
 
 class CsvData:
@@ -61,10 +65,12 @@ class PcapData:
 
 
 class Analyzer:
-    def __init__(self, csv_name, pcap_name, count, test_type):
+    def __init__(
+        self, paths: PathGenerator, count, test_type, ascii_output, save_output
+    ):
         self.count = count
-        self.csv_data = CsvData(csv_name)
-        self.pcap_data = PcapData(pcap_name, test_type)
+        self.csv_data = CsvData(paths.csv())
+        self.pcap_data = PcapData(paths.pcap(), test_type)
 
         graphs = [
             self.ordering_pie_chart,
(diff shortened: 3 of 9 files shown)


xray/.gitignore
```diff
@@ -0,0 +1 @@
+results/
```
mathiaspeters, line 1: I've been adding other xray-related ignores to the top level `.gitignore` so far. I like this idea of having them in the xray folder, but it would make it inconsistent. I'll leave it to you to decide if you move this to the top-level or if you move the other xray-related ones here, but I would like to see it fixed

xray/README.md
```diff
@@ -69,6 +69,12 @@ The application is executed with the `run.py` script. I takes some arguments, al
 
 - **--nobuild-xray**: whether or not to build `xray` before running. Default is to build `xray`
 
+- **--save-output**: save the analysis charts in `results` folder.
+
+- **--ascii**: output the analysis chart as text graph.
+
+- **--disable-drop-privileges**: pass the `disable-drop-privileges` flag to `neptun-cli` and `boringtun-cli`.
```
mathiaspeters, line 76: Is this something that needs to be passed through as a cli arg or can we just pass this unconditionally to neptun/boringtun?

xray/analyze.py
```diff
@@ -187,7 +213,7 @@ def packet_funnel(self, ax):
             f"Recv ({recv})",
         ]
         values = [self.count, before_wg, after_wg, recv]
-        plt.bar(categories, values, color="blue", width=0.4)
+        ax.bar(categories, values, color="blue", width=0.4)
```
mathiaspeters, line 216: Nice catch!

xray/analyze.py
```diff
@@ -84,6 +86,30 @@ def __init__(self, csv_name, pcap_name, count, test_type):
             fn(ax[i, 1])
 
         plt.show()
+        if save_output:
+            plt.savefig(test_path + ".png")
```
mathiaspeters, line 90: Have you tried running this locally with passing `--ascii`? I think I read somewhere that `savefig` needs to be called before `show` because `show` will clear the plot, leaving nothing for `savefig` to save. I saw in the artifacts that it's working though, which is weird. But maybe move the `if save_output` to be before the `show`

xray/run.py
```diff
@@ -13,16 +13,16 @@
...
     return (run.stdout, run.stderr)
 
 
 def get_csv_name(wg, test_type, count):
     return f"results/xray_metrics_{wg.lower()}_{test_type}_{count}.csv"
 
 
-def get_pcap_name(wg, test_type, count):
-    return f"results/{WG_IFC_NAME}_{wg.lower()}_{test_type}_{count}.pcap"
+def get_test_path(wg, test_type, count):
```
mathiaspeters, line 24: I think I would prefer to have a separate function for getting the output files you're creating for now. The naming scheme I started here is not reusable enough (the images you output would be named `xraywg1*` which is the name of the interface). I think a bit of an overhaul of these path getters would be good and then we could centralize on a single function for this

xray/run.py
```diff
@@ -104,6 +110,9 @@ def main():
     parser.add_argument("--count")
     parser.add_argument("--nobuild-neptun", action="store_true")
     parser.add_argument("--nobuild-xray", action="store_true")
+    parser.add_argument("--save-output", action="store_true")
+    parser.add_argument("--disable-drop-privileges", action="store_true")
+    parser.add_argument("--ascii", action="store_true")
```
mathiaspeters, line 115: I think it might be better to have this take a value and then create an enum for it where the options could be `Ascii` and `Graphical` (or something like that) for now. I don't see a value in outputting both formats at the same time. It would also make it slightly easier to understand in the analyzer script if it was `if output_type == OutputType.Ascii:` instead of just `if ascii:`

xray/analyze.py
```diff
@@ -84,6 +86,30 @@ def __init__(self, csv_name, pcap_name, count, test_type):
...
+
+        if ascii:
```
mathiaspeters, line 92: Let's make this an if-else so we only show one of the two formats at the same time. And maybe it can be simplified since both branches would have some code in common

xray/README.md
tomasz-grz: Good question, I guess it depends if we would ever want to run it without `disable-drop-privileges` and how it impacts the tests.

xray/.gitignore
tomasz-grz: Good point! 💯

.github/workflows/tests.yml
```diff
@@ -49,3 +49,35 @@ jobs:
...
+          sudo apt update
+          sudo apt install -y wireguard wireguard-go
+          python -m pip install --upgrade pip
+          pip install pipenv
+          pipenv install
+      - name: Execute xray
+        working-directory: xray
+        run: |
+          pipenv run python run.py --wg native --ascii --save-output
+          pipenv run python run.py --wg neptun --ascii --save-output --disable-drop-privileges
```
mathiaspeters, line 70: We should also run these with `--test-type plaintext`, and we probably want to have a bigger package count too, maybe 10k

xray/analyze.py
tomasz-grz: hmm yeah maybe because it doesn't show the UI on the CI.

xray/run.py
tomasz-grz: Are we anticipating to have more than two, `Ascii` and `Graphical` output types to justify the additional boiler plate code?

tomasz-grz: I thought it was quite convenient to have a more detailed graphical view available to investigate the result, but if we don't see cases that that will be useful I will remove it.

.github/workflows/tests.yml
tomasz-grz: Good idea!

xray/.gitignore
tomasz-grz: Moved to top level `.gitignore`

xray/analyze.py
tomasz-grz: Swapped the order

xray/run.py
tomasz-grz: Good idea, refactored this into a separate module

xray/analyze.py
tomasz-grz: Converted into an if-else case

xray/paths.py
```diff
@@ -0,0 +1,24 @@
+class PathGenerator:
+
+    def __init__(self, wg_ifc_name: str, wg: str, test_type: str, count: int):
+        self.wg_ifc_name = wg_ifc_name
+        self.wg = wg
+        self.test_type = test_type
+        self.count = count
+
+    def base_path(self) -> str:
+        return f"results/{self.wg_ifc_name}_{self.wg.lower()}_{self.test_type}_{self.count}"
```
mathiaspeters, line 10: Let's make this `f"results/xray_{self.wg.lower()}_{self.test_type.lower()}_{self.count}"`. 

I started naming the pcaps after the interface but the pcaps collect on all interfaces, and regardless of that, the output of the analysis should not be named after the iterface. I like having a more generic naming here, which your changes go a long way towards fixing, but I think this is one last bit that needs to be changed. And along with that, the `csv(self)` function should return `f"{self.base_path()}.csv`

A minor nit for this function too: `base_path` should be private and the way that is done in python would be to name it `_base_path`

xray/README.md
mathiaspeters: Let's leave it like this for now and we can investigate it at a later point if necessary

xray/run.py
mathiaspeters: You're right, it's probably overkill for now, we can keep this as it is :smile:

xray/paths.py
ghztomash: Good idea, fixed this

mathiaspeters:
+1 :rocket:

dfetti (approved):
+1.0

PR #17, opened 2024-12-19
Fix hanging on SIGINT

jjanowsk:
Ask other worker threads to yield the device lock when using try_writeable. Otherwise it will hang indefinitely since other threads constantly keep the read lock


dfetti (approved):
+2.0

PR #18, opened 2025-01-06
Add bidirectional testing to xray

mathiaspeters:
This PR consists of two commits that are best reviewed separately.

The first commit is a cleanup commit that does a few things:
1. Adds proper types for wireguard adapter and test type in xray
2. Makes clap aware of the wg and test types types so it doesn't have to be manually parsed
3. Adds `TestType` class in `run.py` 
4. Create `types.rs` in xray and move a bunch of xray-specific types to that file
5. Remove a png that should have been removed in a previous PR

The second commit actually implements bidirectional testing, which consists of a few steps:
1. Add `Bidir` variant to both python and rust `TestType`
2. Update some prints to remove `[Plaintext]` and `[Crypto]` since for the purposes of that print we care about the total number of packets sent, not really about which direction it's being sent
3. Update how the pcap parser determines if a packet is before or after wg


xray/src/event_loop.rs
```diff
@@ -153,12 +155,10 @@ impl EventLoop {
             }
             RecvType::Data { length: bytes_read } => {
                 if bytes_read == Packet::send_size() {
-                    if self.recv_counter > 0
-                        && self.recv_counter % (self.cli_args.packet_count / 10) == 0
-                    {
-                        println!("[Crypto] Received {} packets", self.recv_counter);
-                    }
                     self.recv_counter += 1;
```
tomasz-grz, line 158: What do you think about moving this below the `if` statement so that we don't have to do `self.recv_counter - 1`

xray/src/utils.rs
```diff
@@ -120,9 +56,12 @@ pub async fn configure_wg(
     }
 
     match adapter_type {
-        "native" | "boringtun" => configure_native_wg(wg_name, wg_keys, peer_keys, wg_port),
-        "wggo" | "neptun" => configure_userspace_wg(wg_name, wg_keys, peer_keys, wg_port).await,
-        _ => Err(XRayError::UnknownAdapter(adapter_type.to_owned())),
+        Wg::LinuxNative | Wg::BoringTun => {
```
tomasz-grz, line 59: Why is `BoringTun` also not part of userspace?

xray/src/main.rs
```diff
@@ -150,21 +123,41 @@ async fn main() -> EyreResult<()> {
...
+                    .await?
+            }
+            TestType::Bidir => {
+                cmd_tx
+                    .send(TestCmd::SendPlaintext {
+                        dst: CRYPTO_ADDR,
+                        send_index: i as u64,
+                    })
+                    .await?
+            }
```
tomasz-grz, line 160: nit: I would personally find a single match arm with an inner if statement more intuitive to read.

xray/src/main.rs
```diff
@@ -39,48 +38,22 @@ const PLAINTEXT_ADDR: SocketAddrV4 = SocketAddrV4::new(*WG_ADDR.ip(), PLAINTEXT_
...
+            .map(|s| s.parse::<Wg>().unwrap()),
+    )]
+    wg: Wg,
+    #[arg(
+        long,
+        default_value_t = TestType::Crypto,
+        value_parser = clap::builder::PossibleValuesParser::new(["crypto", "plaintext", "bidir"])
+            .map(|s| s.parse::<TestType>().unwrap()),
+    )]
+    test_type: TestType,
```
tomasz-grz, line 56: Nice!

tomasz-grz (commented):
Great work! A couple of questions from my side

tomasz-grz:
Would it be beneficial to already add the bidirectional tests to the CI job? (or replacing the other two if it tests the same cases as plaintext + crypto). 
Speaking of those test results, why do we have so many dropped packets?

xray/src/event_loop.rs
mathiaspeters: We can do that, but then we'd do `self.recv_counter + 1` in the `println!` inside the if-statement. I'm okay doing it either way but there will be arithmetic done either way

xray/src/main.rs
mathiaspeters: Fixed. I also cleaned it up by having the match return a `TestCmd` and then send it after the match

mathiaspeters:
> Would it be beneficial to already add the bidirectional tests to the CI job? (or replacing the other two if it tests the same cases as plaintext + crypto).
> Speaking of those test results, why do we have so many dropped packets?

Adding it to the CI is a good idea! 

Most of the dropped packets are likely because of overhead added by xray itself, like processing packets after they arrive, or at least that's the case when sending plaintext->crypto. That will be fixed in a separate effort though, and we of course won't know how much of the problem that would fix

xray/src/utils.rs
tomasz-grz: Hmm I maybe it makes sense to remove it, also wggo if it's not really supported at the moment. Unless we are planning to add support for it shortly.

xray/src/event_loop.rs
tomasz-grz: If `self.recv_counter + 1` would be only for correct visual representation, then I would prefer it inside `println!`. It's just slight less cognitive load when reading `if (self.recv_counter - 1) % (self.cli_args.packet_count / 10) == 0`.

Since it's only a for the print statement consider this just a nit. 😅

tomasz-grz: On the other hand, Is `self.recv_counter` an index?

In the logs of the test run you can see.
```
1 packets recevied
1001 packets recevied
2001 packets recevied
```

where as before this PR:
```
[Plaintext] Received 1000 packets
[Plaintext] Received 2000 packets
[Plaintext] Received 3000 packets
```

xray/src/main.rs
tomasz-grz: Nice, great work cleaning it up. 
By single match arm, I think i meant something along the lines of:
```rust
            TestType::Bidir => {
                if i % 2 == 0 {
                    TestCmd::SendEncrypted {
                        sock_dst: WG_ADDR,
                        packet_dst: PLAINTEXT_ADDR,
                        send_index: i as u64,
                    }
                } else {
                    TestCmd::SendPlaintext {
                        dst: CRYPTO_ADDR,
                        send_index: i as u64,
                    }
                }
            }
```
Either way, just a nit. 👌

xray/src/event_loop.rs
mathiaspeters: It's a good catch. I think I was mostly testing this with 10 packets sent and wanted to get that working properly, but then didn't test multiple packets. Now it should be fixed for both cases

xray/src/utils.rs
mathiaspeters: I removed boringtun, but leaving WireguardGo since that's a target we want to support. Something went wrong at some point where it no longer works in both directions, but that's something we want to fix, just not in this PR

xray/src/main.rs
mathiaspeters: I did that first but then had some git issues and had to redo it, but forgot to do what you actually asked for :joy: It's fixed now, for real

tomasz-grz: Awesome, thanks!

tomasz-grz (approved):
+2.0 🚢 📦 it

PR #19, opened 2025-01-10
break tx into smaller tasks

Hasan6979:


neptun/src/device/mod.rs
```diff
@@ -1123,54 +1156,89 @@ impl Device {
...
                 for _ in 0..MAX_ITR {
-                    let src = match iface.read(&mut t.src_buf[..mtu]) {
-                        Ok(src) => src,
-                        Err(Error::IfaceRead(e)) => {
-                            let ek = e.kind();
-                            if ek == io::ErrorKind::Interrupted || ek == io::ErrorKind::WouldBlock {
-                                break;
+                    let block = unsafe { TX_RING_BUFFER.get_next() };
+                    let mut element = block.lock();
+                    if element.is_element_free {
```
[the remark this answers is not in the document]
Hasan6979: Yes 💯  but we would want to have a very large `MAX_ITR` to avoid `epoll` calls, and having a very large `len(TX_RING_BUFFER)` won't be possible so a balance is needed that ring buffer size is large enough that our packet drops are reduced as much as possible.

neptun/src/device/ring_buffers.rs
```diff
@@ -0,0 +1,50 @@
...
+use once_cell::sync::Lazy;
+use parking_lot::Mutex;
+use std::sync::Arc;
+const UDP_SIZE: usize = 1600;
+
+pub const RB_SIZE: usize = 500;
+
+pub struct RingBuffer<T> {
+    pub ring_buffer: Vec<T>,
+    iter: Mutex<usize>,
```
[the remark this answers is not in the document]
Hasan6979: Ring buffer has been removed. Copying the data to the new thread resulted in no performance drop.

Hasan6979:
UDP tests
Base NepTUN       | 51% packets dropped  @ 495 Mbits/sec 
Current NepTUN  |  15% packets dropped @ 853 Mbits/sec 

TCP tests
Base NepTUN | 355 Mbits/sec
Current NepTUN | 506 Mbits/sec

details https://github.com/NordSecurity/NepTUN/actions/runs/15042748104/job/42278110912?pr=19

matislovas:
+1

neptun/src/device/mod.rs
```diff
@@ -1088,13 +1149,13 @@ impl Device {
                 // * Send encapsulated packet to the peer's endpoint
                 let mtu = d.mtu.load(Ordering::Relaxed);
 
-                let udp4 = d.udp4.as_ref().expect("Not connected");
-                let udp6 = d.udp6.as_ref().expect("Not connected");
-
                 let peers = &d.peers_by_ip;
                 for _ in 0..MAX_ITR {
```
stalowyjez, line 1153: Nit: If we make this `MAX_ITR` so large, maybe it would be better to just use `loop`?

neptun/src/device/mod.rs
```diff
@@ -1123,16 +1186,56 @@ impl Device {
...
+    close_chan: Receiver<()>,
+    udp4: Arc<socket2::Socket>,
+    udp6: Arc<socket2::Socket>,
+    firewall_process_outbound_callback: Option<
+        Arc<dyn Fn(&[u8; 32], &[u8], &mut dyn std::io::Write) -> bool + Send + Sync>,
+    >,
+) {
+    loop {
+        crossbeam_channel::select! {
+            recv(network_rx) -> m => {
```
stalowyjez, line 1218: Nit: Maybe you could just use `element` instead of `m`?

neptun/src/device/mod.rs
```diff
@@ -1123,16 +1186,56 @@ impl Device {
...
+                    let peer = if let Some(p) = element.peer.clone() {
+                        p
+                    } else {
+                        tracing::error!("Empty peer");
+                        continue;
+                    };
+
+                    if let Some(callback) = &firewall_process_outbound_callback {
+                        if let Some(i) = element.iface {
+                        if !callback(&peer.public_key.0, &element.data[WG_HEADER_OFFSET..element.buf_len + WG_HEADER_OFFSET], &mut i.as_ref()) {
```
stalowyjez, line 1230: Nit: Hmm, so you have this `let len = element.buf_len;` maybe you could use it here as well?

neptun/src/device/mod.rs
```diff
@@ -1123,16 +1186,56 @@ impl Device {
...
                             continue;
                         }
                     }
+                    }
```
stalowyjez, line 1234: Nit: it looks to me like some formatting issue

neptun/src/noise/session.rs
```diff
@@ -28,7 +28,7 @@ impl std::fmt::Debug for Session {
 }
 
 /// Where encrypted data resides in a data packet
-const DATA_OFFSET: usize = 16;
+pub const DATA_OFFSET: usize = 16;
```
stalowyjez, line 31: Nit: Hmm, maybe it could be `pub(crate)`?

neptun/src/noise/session.rs
```diff
@@ -194,8 +194,8 @@ impl Session {
     /// src - an IP packet from the interface
```
stalowyjez, line 194: It looks that this description now doesn't match with the function signature

neptun/src/noise/session.rs
```diff
@@ -194,8 +194,8 @@ impl Session {
...
     /// dst - pre-allocated space to hold the encapsulating UDP packet to send over the network
     /// returns the size of the formatted packet
-    pub(super) fn format_packet_data<'a>(&self, src: &[u8], dst: &'a mut [u8]) -> &'a mut [u8] {
-        if dst.len() < src.len() + super::DATA_OVERHEAD_SZ {
+    pub(super) fn format_packet_data<'a>(&self, src_len: usize, dst: &'a mut [u8]) -> &'a mut [u8] {
```
stalowyjez, line 197: It looks to me a bit that this function works on some header now, maybe it would be better to write it explicitely

stalowyjez (approved):
Looks good, just a few nits, +1

neptun/src/noise/session.rs
Hasan6979: What do you mean by "works on some header" ? The main functionality of this function hasn't changed. Earlier the data was being copied to a buffer then encrypted in place. Now the copy part isn't there.

neptun/src/device/mod.rs
Hasan6979: That constant is used by `udp_handler` as well. I reverted it to `100` as it was originally.

neptun/src/noise/session.rs
stalowyjez: There are still this names `src` and `dst`, so I assumed that the data is not in `dst` yet - after the description is changed it became a bit clearer what happens here to me, still, I think `payload_len` and `unformatted_packet` or sth similar would be more intuitive as currently the data is already in `dst` which is a bit misleading imho

Hasan6979: Makes sense!

stalowyjez (approved):
Lgtm

PR #21, opened 2025-01-23
Use `cargo xtask` for xray

mathiaspeters:
xray was initially created with a python run script. Since then, `cargo xtask` was introduced for some other performance testing, which is a nicer solution than using python scripts. This PR replaces `run.py` with a `cargo xtask` setup.

Notes:
- This change required `analyze.py` to become a standalone script with its own argument parser
- The `count` parameter was renamed to `packet_count` to be more explicit


xray/README.md
```diff
@@ -56,13 +56,13 @@ Supported adapter types:
...
 
 - **--wg <adapter type>**: which adapter type to use, accepts one of `neptun`, `native` and `wggo`. Default is `neptun`
 
-- **--test_type <test type>**: which kind of test to run, accepts `crypto` (to send packets from the crypto socket to the plaintext socket), `plaintext` (to send packets from the plaintext socket to the crypto socket), and `bidir` (to send packets in both directions). Default is `crypto`
+- **--test-type <test type>**: which kind of test to run, accepts `crypto` (to send packets from the crypto socket to the plaintext socket), `plaintext` (to send packets from the plaintext socket to the crypto socket), and `bidir` (to send packets in both directions). Default is `crypto`
 
-- **--count <number of packets>**: How many packets to send. Default is 10
+- **--packet-count <number of packets>**: How many packets to send. Default is 10
 
 - **--nobuild-neptun**: whether or not to build `neptun-cli` before running. Only relevant when the adaptert type is `neptun`. Default is to build `neptun`
```
tomaszklak, line 67: Is this option still possible? I don't see it in the `xray/src/lib.rs`.

tomaszklak, line 67: Maybe it would be better to move this parameter description to the `xray/src/lib.rs` so that it's 'attached' to the actual agrument parsing, is shown as a help when running the task and it's easier to keep in sync.

xray/src/path_generator.rs
```diff
@@ -0,0 +1,59 @@
...
+    pub fn analyze_script(&self) -> PathBuf {
+        self.base_dir.join("xray/analyze.py")
+    }
+
+    pub fn results_dir(&self) -> PathBuf {
+        self.base_dir.join("xray/results")
+    }
+
+    pub fn binary_dir(&self) -> PathBuf {
+        self.base_dir.join("target/release/")
```
tomaszklak, line 33: nit: the trailing `/` is not needed

xtask/src/xray.rs
```diff
@@ -0,0 +1,221 @@
...
+        let path_builder = PathGenerator::new(self.wg, self.test_type, self.packet_count);
+
+        let results_dir = path_builder.results_dir();
+        if !results_dir.exists() {
+            std::fs::create_dir_all(results_dir).expect("Failed to create results folder");
+        }
+        let _ = std::fs::remove_file(path_builder.csv());
+        let _ = std::fs::remove_file(path_builder.pcap());
+        let _ = std::fs::remove_file(path_builder.png());
+        let _ = std::fs::remove_file(path_builder.txt());
```
tomaszklak, line 95: If the goal is to have an empty `path_builder.results_dir()` I think it would be simpler (and more correct) to first `remove_dir_all(path_builder.results_dir())` followed by `create_dir_all`.

xtask/src/xray.rs
```diff
@@ -0,0 +1,221 @@
...
+
+        let xray_res = self.run_xray(&sh, &path_builder);
+        if let Err(e) = &xray_res {
+            println!("{e}");
```
tomaszklak, line 99: nit: printing errors is usually done on the stderr, so `eprintln` could be used here.

xtask/src/xray.rs
```diff
@@ -0,0 +1,221 @@
...
+        }
+
+        let _ = self.stop_tcpdump(&sh);
+        if let Err(e) = self.destroy_wg_adapter(&sh) {
+            println!("{e}");
```
tomaszklak, line 104: nit: printing errors is usually done on the stderr, so `eprintln` could be used here.

xtask/src/xray.rs
```diff
@@ -0,0 +1,221 @@
...
+            let packet_count = self.packet_count.to_string();
+            let base_path = path_builder.base();
+            let analyze_res = match (self.save_output, self.ascii) {
+                (true, true) => run_cmd!(sh, "{analyze_script} --packet-count {packet_count} --base-path {base_path} --save-output --ascii"),
+                (true, false) => run_cmd!(sh, "{analyze_script} --packet-count {packet_count} --base-path {base_path} --save-output"),
+                (false, true) => run_cmd!(sh, "{analyze_script} --packet-count {packet_count} --base-path {base_path} --ascii"),
+                (false, false) => run_cmd!(sh, "{analyze_script} --packet-count {packet_count} --base-path {base_path}"),
+            };
+            if let Err(e) = analyze_res {
+                println!("{e}");
```
tomaszklak, line 118: nit: printing errors is usually done on the stderr, so `eprintln` could be used here.

xtask/src/xray.rs
```diff
@@ -0,0 +1,221 @@
...
+    fn build(&self, sh: &Shell) -> Result<()> {
+        let should_build_neptun = matches!(self.wg, Wg::NepTUN) && !self.nobuild_neptun;
+        let should_build_xray = !self.nobuild_xray;
+
+        match (should_build_neptun, should_build_xray) {
+            (true, true) => run_cmd!(sh, "cargo build --release -p neptun-cli -p xray")?,
+            (true, false) => run_cmd!(sh, "cargo build --release -p neptun-cli")?,
+            (false, true) => run_cmd!(sh, "cargo build --release -p xray")?,
+            (false, false) => {}
+        }
```
tomaszklak, line 150: nit: shouldn't cargo be smart enough to know what needs to be build? If that's the case it should be enough to call `cargo build --release -p neptune-cli -p xray` if any of the two flags is enabled.

xtask/src/xray.rs
```diff
@@ -0,0 +1,221 @@
...
+    fn start_tcpdump(&self, sh: &Shell, pcap_path: PathBuf) -> JoinHandle<()> {
+        let sh = sh.clone();
+        let packet_filter =
+            format!("udp and (port {WG_PORT} or port {PLAINTEXT_PORT} or port {CRYPTO_PORT})");
+        let handle = std::thread::spawn(move || {
+            run_cmd!(sh, "sudo tcpdump -ni any -w {pcap_path} {packet_filter}")
+                .expect("Failed to start tcpdump");
+        });
+        // Small sleep to make sure tcpdump has started before we continue
+        std::thread::sleep(std::time::Duration::from_secs(1));
```
tomaszklak, line 205: Is this meant to be used in some automated tests? If so, this looks like a source of future flakiness.

xtask/src/xray.rs
```diff
@@ -0,0 +1,221 @@
...
+        handle
+    }
+
+    fn stop_tcpdump(&self, sh: &Shell) -> Result<()> {
+        run_cmd!(sh, "sudo killall -w tcpdump")?;
```
tomaszklak, line 210: nit: I think it's better to use long versions of flags in scripts, so that it's easier for future readers to understand the meaning without looking it up in the manual. There this means that the `-w` should be replaced with `--wait`. This applies to numerous other places in this file.

xtask/src/xray.rs
mathiaspeters: The goal is to remove all files for the current configuration. For example, if I ran it with `neptun` and got some results and then run with `native` I don't want to remove the results for `neptun` since they are non-overlapping and could/would be used for manual comparison

xray/README.md
mathiaspeters: It's a little bit confusing at the moment but yes it's still there, but it's in `xtask/src/xray.rs`

.github/workflows/tests.yml
```diff
@@ -70,12 +73,12 @@ jobs:
       - name: Execute xray
         working-directory: xray
         run: |
-          pipenv run python run.py --wg native --ascii --save-output --test-type crypto --count 10000
-          pipenv run python run.py --wg native --ascii --save-output --test-type plaintext --count 10000
-          pipenv run python run.py --wg native --ascii --save-output --test-type bidir --count 10000
-          pipenv run python run.py --wg neptun --ascii --save-output --test-type crypto --count 10000 --disable-drop-privileges
-          pipenv run python run.py --wg neptun --ascii --save-output --test-type plaintext --count 10000 --disable-drop-privileges
-          pipenv run python run.py --wg neptun --ascii --save-output --test-type bidir --count 10000 --disable-drop-privileges
+          pipenv run cargo xtask xray --wg native --ascii --save-output --test-type crypto --packet-count 10000
```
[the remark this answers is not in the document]
mathiaspeters: Unfortunately yes. `analyze.py` is still there and that relies on python and pipenv. I have some ideas for a future PR though

xray/README.md
mathiaspeters: Good idea. Fixed

xray/src/path_generator.rs
mathiaspeters: In this case it kinda was due to how it's used, but I removed it from here to make this file more consistent and added the `/` in the use sites since that also makes those places clearer

xtask/src/xray.rs
mathiaspeters: Fixed

xtask/src/xray.rs
mathiaspeters: Fixed

xtask/src/xray.rs
mathiaspeters: Fixed

xtask/src/xray.rs
mathiaspeters: This separation was initially added when the builds were separated (i.e. `cargo build -p neptun-cli` before starting neptun and then a separate `cargo build -p xray` before starting the xray binary). With the separate builds, cargo wasn't smart enough to used the cached build so the `nobuild-neptun` and `nobuild-xray` flags were introduced so you didn't have to wait for two separate builds every time. Then the cause for the constant rebuilds was discovered and the fix was to combine both builds in a single statement and cleaned up here. Since building like this will much better use cached artifacts, the separation of the `nobuild-<something>` is not really necessary anymore so I replaced it with a single `nobuild` argument to build or not build both crates

xtask/src/xray.rs
mathiaspeters: We are running it in CI but so far it's mostly been used for local runs. I don't like this sleep either but I can't think of a better way to check that `tcpdump` is ready

xtask/src/xray.rs
mathiaspeters: I fixed this where possible

xray/src/utils.rs
```diff
@@ -29,7 +29,7 @@ pub fn run_command(cmd: String) -> Result<String, String> {
     }
 }
 
-pub fn write_to_csv(name: &str, packets: &[Packet]) -> XRayResult<()> {
+pub fn write_to_csv(name: PathBuf, packets: &[Packet]) -> XRayResult<()> {
```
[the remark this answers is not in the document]
mathiaspeters: I don't think it makes too much difference in practice just based on how the paths are used, but it's a good idea in general so it's fixed

xtask/src/xray.rs
tomaszklak: Afaics, even with `-w` tcpdump prints some stuff on the stderr:
```
$ sudo tcpdump -i any -w /dev/null
tcpdump: data link type PKTAP
tcpdump: listening on any, link-type PKTAP (Apple DLT_PKTAP), snapshot length 524288 bytes
```
So we could capture stderr and wait till it contains `listening on ` string.

tomaszklak (dismissed):
+1 (after https://github.com/NordSecurity/NepTUN/pull/21#discussion_r1930224510 is resolved)

.github/workflows/tests.yml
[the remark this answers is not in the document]
tomasz-grz: Oh right, somehow missed that

xtask/src/xray.rs
mathiaspeters: Ticket and `//TODO` created

xray/src/pcap.rs
```diff
@@ -18,7 +18,7 @@ pub struct PcapPacket {
     pub send_index: u64,
 }
 
-pub fn process_pcap(pcap_path: &str, mut tunn: Tunn) -> XRayResult<Vec<PcapPacket>> {
+pub fn process_pcap(pcap_path: PathBuf, mut tunn: Tunn) -> XRayResult<Vec<PcapPacket>> {
```
tomasz-grz, line 21: I feel that maybe we should also change this one to keep them consistent with `write_to_csv`

tomasz-grz (commented):
Another small nit, otherwise +1

tomasz-grz (approved):

PR #22, opened 2025-02-03
Make Persistent-Keepalive trigger handshakes

LukasPukenis:
There are two cases where NepTUN might hit `ConnectionExpired` error and not proceed further in the `update_timers()` function. This has been reproduced with tcli by establishing VPN connection, suspending the process, waiting for 540seconds, resuming and observing that the state is `connecting`.

It should happen if there are no persistent-keepalives set, however in our case we have them, so the fix revolves around checking for it.

This was checked with tcli which emits keepalive package and in turn a handshake attempt afte resuming.

Both fixes were unit tested as well.

### DIFF
--- a/neptun/src/noise/mod.rs
+++ b/neptun/src/noise/mod.rs
@@ -675,8 +675,11 @@ impl Tunn {
 
 #[cfg(test)]
 mod tests {
+    use crate::noise::timers::{
+        REJECT_AFTER_TIME, REKEY_AFTER_TIME, REKEY_ATTEMPT_TIME, REKEY_TIMEOUT,
+    };
     #[cfg(feature = "mock-instant")]
-    use crate::noise::timers::{REKEY_AFTER_TIME, REKEY_TIMEOUT};
+    use mock_instant::MockClock;
 
     use super::*;
     use rand_core::{OsRng, RngCore};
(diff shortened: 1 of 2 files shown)


neptun/Cargo.toml
```diff
@@ -52,6 +52,7 @@ tracing-subscriber = "0.3"
 criterion = { version = "0.5", features = ["html_reports"] }
 rand = "0.8"
 tokio = { version = ">=1.22", features = ["rt", "net", "sync", "macros"] }
+test-case = "*"
```
mathiaspeters, line 55: Let's set this to `3` or `3.3`

neptun/src/noise/mod.rs
```diff
@@ -818,6 +819,54 @@ mod tests {
         assert!(matches!(their_tun.update_timers(&mut []), TunnResult::Done));
     }
 
+    #[cfg(feature = "mock-instant")]
+    #[test_case(false; "Persistent-Keepalive not set, should not result in handshakes after expiration")]
+    #[test_case(true; "Persistent-Keepalive is set, should result in handshakes after expiration")]
+    fn after_3x_REJECT_AFTER_TIME(nonzero_ka: bool) {
```
mathiaspeters, line 825: Nit: I think the name `nonzero_ka` is a bit cryptic. How about something like `set_keepalive` instead? Or `should_set_keepalive` to be even more explicit

neptun/Cargo.toml
LukasPukenis: Good point, however I removed it entirely instead as testcases became difficult to debug even though they have a lot of similarity :)

neptun/src/noise/mod.rs
LukasPukenis: The point for it was to indicate that any keepalive value would do while it's non-zero, but instead I just hardcoded `25` seconds in most cases which makes it simpler IMO

mathiaspeters (commented):
+1

ThrasherLT (approved):
+1

PR #23, opened 2025-02-10
Add notice about testing keys

Jauler:
NepTUN does some performance measurements, which requires to setup real connections between peers. For that some test keys has been generated, which are not used in any real deployment. Add a note explicitly stating this fact to avoid confusion.


LukasPukenis (approved):
+2

PR #24, opened 2025-03-10
Downgrade the message indicating that syscall was interrupted from error to trace

tomaszklak:
I tried to test/reproduce it in few different ways, like sending SIGUSR1 to the threads waiting in `epoll_wait`, but it doesn't result in the reported message, but it kill the whole process.


LukasPukenis (approved):
Despite the inability to reproduce it, the fix itself looks like what I would expect 
+2

PR #25, opened 2025-03-20
Change set_iface to take into account multiple threads running event_loop

tomaszklak:
When requested, NepTUN starts multiple threads reading from the same Tun interface. This means that all of those threads need to somehow be modified when we change the Tun from which reading should be done.

This fix stops all 'old' threads (by closing the file descriptors from which they read), and starts fresh threads opening fresh file descriptors from the specified Tun.


gytsto (commented):
+1

matislovas (approved):
+1

PR #27, opened 2025-04-03
Replace manual ioctl calls with nix crate

jjanowsk:
ifr in new_from_fd was not marked mut. This was a bug since ioctl call
modified this variable. This lead to optimizer in 1.85.0 version of Rust
to optimize away the access to this variable after ioctl call and just
storing hardcoded 0 in the flags variable.
This commit fixes it and additionally:
1. Uses ioctl macro from nix crate
2. Removes unnecesary struct definitions which are available in libc
   crate


LukasPukenis (commented):
+1

neptun/src/device/tun_linux.rs
```diff
@@ -3,46 +3,33 @@
...
 }
 
-#[repr(C)]
-pub struct ifreq {
-    ifr_name: [c_uchar; IFNAMSIZ],
-    ifr_ifru: IfrIfru,
+#[cfg(target_os = "android")]
+mod tun_interface_flags {
+    use super::*;
+    const TUNGETIFF: u64 = 0x8004_54D2;
```
jjanowsk, line 27: I've opened a https://github.com/rust-lang/libc/pull/4379 so hopefully this will soon be removed

LukasPukenis: Very nice, I was googling for this to see if maybe it's already defined somewhere when reviewing :)

neptun/src/device/tun_linux.rs
```diff
@@ -98,9 +88,9 @@ impl TunSocket {
                 return Err(Error::InvalidTunnelName);
             }
 
-            ifr.ifr_name[..iface_name.len()].copy_from_slice(iface_name);
+            ifr.ifr_name[..iface_name.len()].copy_from_slice(&iface_name);
```
tomaszklak, line 91: nit: it looks like we are allocating temporary slice just to copy out of it here. Maybe it would be simple to avoid that temporary?

tomaszklak (approved):
+1

tomaszklak (approved):

PR #31, opened 2025-04-23
Refactor setting socket buffer prints friendlier

Hasan6979:
Remove unsafe in setting socket buffer using nix sockets instead of libc

### DIFF
--- a/neptun/src/device/mod.rs
+++ b/neptun/src/device/mod.rs
@@ -26,11 +26,12 @@ pub mod tun;
 #[path = "tun_linux.rs"]
 pub mod tun;
 
+use nix::sys::socket as NixSocket;
 use std::collections::HashMap;
 use std::io::{self, BufReader, BufWriter};
 use std::mem::{swap, MaybeUninit};
 use std::net::{IpAddr, Ipv4Addr, Ipv6Addr, SocketAddr, SocketAddrV4, SocketAddrV6};
-use std::os::fd::RawFd;
+use std::os::fd::{AsFd, BorrowedFd, RawFd};
 #[cfg(not(target_os = "windows"))]
 use std::os::unix::io::AsRawFd;
 use std::sync::atomic::{AtomicUsize, Ordering};
@@ -47,7 +48,6 @@ use peer::{AllowedIP, Peer};
 use poll::{EventPoll, EventRef, WaitResult};
 use rand_core::{OsRng, RngCore};
 use socket2::{Domain, Protocol, Type};
-use tracing::{debug, warn};
 use tun::TunSocket;
 
 use dev_lock::{Lock, LockReadGuard};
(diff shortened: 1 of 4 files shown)


neptun/src/device/mod.rs
```diff
@@ -449,9 +449,16 @@ fn modify_skt_buffer_size(socket: i32, buffer_size: u32) {
...
+            libc::SO_RCVBUF => "SO_RCVBUF",
+            libc::SO_SNDBUF => "SO_SNDBUF",
+            _ => "Unknown",
+        };
+
         match res {
-            0 => tracing::debug!("Socket buffer {buffer} set"),
-            _ => tracing::error!("Socket buffer {buffer} failed with {res}"),
+            0 => tracing::info!("Socket buffer {buffer_name} set with value {buffer_size}"),
+            _ => tracing::warn!("Socket buffer {buffer_name} failed with {res}"),
```
mathiaspeters, line 461: Do we know if `res` has some nice type mapping in the libc crate so that we can print something more informative than just an integer?

Hasan6979: There is no nice direct mapping, but we could simply do an exhaustive match. There are only 7 possible errors that can be thrown as given by https://man7.org/linux/man-pages/man3/setsockopt.3p.html.
Wdyt ?

mathiaspeters: Sounds good to me

neptun/src/device/mod.rs
```diff
@@ -449,9 +449,26 @@ fn modify_skt_buffer_size(socket: i32, buffer_size: u32) {
...
             )
         };
+
+        let buffer_name = match buffer {
+            libc::SO_RCVBUF => "SO_RCVBUF",
+            libc::SO_SNDBUF => "SO_SNDBUF",
+            _ => "Unknown",
+        };
+
         match res {
```
tomasz-grz, line 459: does `setsockopt()` set `perror()`? Or maybe we can get `strerror()` or `strerrorname_np()`?

neptun/src/device/mod.rs
```diff
@@ -449,9 +449,26 @@ fn modify_skt_buffer_size(socket: i32, buffer_size: u32) {
...
-            0 => tracing::debug!("Socket buffer {buffer} set"),
-            _ => tracing::error!("Socket buffer {buffer} failed with {res}"),
+            0 => tracing::info!("Socket buffer {buffer_name} set with value {buffer_size}"),
```
tomasz-grz, line 460: Actually should this function return some sort of result?

tomasz-grz (commented):
Would personally like to see at least a short PR description regarding the motivation of these changes.

mathiaspeters:
+1

tomasz-grz (commented):
Nice +1

neptun/src/device/mod.rs
```diff
@@ -1295,7 +1297,7 @@ mod tests {
         let _res = socket.bind(&SocketAddrV4::new(Ipv4Addr::UNSPECIFIED, 0).into());
 
         let mut buffer_size = 11111;
```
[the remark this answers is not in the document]
Hasan6979: Arbitrary number. We can `const` it, but it doesn't matter though.

neptun/src/device/mod.rs
Hasan6979: Since the result is being handled internally within the function, and it isn't critical to other operations even if it fails so I don't think this should return result.

tomasz-grz (approved):

PR #33, opened 2025-05-08
Implement io::Write for TunSocket

jjanowsk:
Add &dyn io::Write parameter to the firewall_process_outbound_callback and pass TunSocket there so that firewall can respond with reject packets.

Additionally TunSocket is stripped from write4/write6 methods which where only needed on darwin platforms. Since the only allowed write to the tun device is the proper IP packet we can read the IP version from the IP header directly in the write method. This way TunSocket can implement io::Write trait.

### DIFF
--- a/neptun/src/device/tun_darwin.rs
+++ b/neptun/src/device/tun_darwin.rs
@@ -4,7 +4,7 @@
 
 use super::Error;
 use libc::*;
-use std::io;
+use std::io::{self, Write};
 use std::mem::size_of;
 use std::mem::size_of_val;
 use std::os::unix::io::{AsRawFd, RawFd};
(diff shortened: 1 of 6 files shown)


neptun/src/device/tun_darwin.rs
```diff
@@ -86,8 +86,24 @@ pub fn parse_utun_name(name: &str) -> Result<u32, Error> {
...
+    fn write(&mut self, src: &[u8]) -> io::Result<usize> {
+        (&*self).write(src)
+    }
+
+    fn flush(&mut self) -> io::Result<()> {
+        (&*self).flush()
+    }
+}
+
+impl Write for &TunSocket {
```
tomaszklak, line 99: Why do we need both implementations?

tomaszklak (commented):
Please add tests for the new functionality.

tomaszklak:
> TunSocket is stripped from write4/write6 methods which where only needed on darwin platforms

Could you clarify - why was it only needed on darwin, and what changed that those methods are no longer needed?

gytsto (commented):
+1

neptun/src/device/tun_darwin.rs
jjanowsk: In theory we do not need impl Write for TunSocket. We do not use it. However it felt strange to not implement it and only implement it for &TunSocket. Unless you're asking why to implement the trait for &TunSocket in the first place. Then the reason is that we want to allow users to use write method even if they do not have mutable reference to TunSocket (and io::Write::write method requires &mut self). However they can always mutably borrow the reference itself. This is a common pattern in e.g. std library and is safe as long as writes to the fd are serialized by the kernel.

jjanowsk:
> > TunSocket is stripped from write4/write6 methods which where only needed on darwin platforms
> 
> Could you clarify - why was it only needed on darwin, and what changed that those methods are no longer needed?

Nothing has changed, I think the original implementor of this interface wanted to play it safe. However the only allowed write to the TUN device is the complete IP packet so we can always get this information from the header. It's only needed on darwin because linux kernel handles it internally the same way we do it now, by checking the first byte if you pass it `IFF_NO_PI` flag during initialization. From tuntap docs:
```
3.2 Frame format:
  If flag IFF_NO_PI is not set each frame format is: 
     Flags [2 bytes]
     Proto [2 bytes]
     Raw protocol(IP, IPv6, etc) frame.
```
Without the flag we would need to provide the same 4 bytes as we do now for darwin

PR #34, opened 2025-05-09
Add code coverage in CI

Hasan6979:
This PR adds code coverage to NepTun repo

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -19,6 +19,26 @@ jobs:
     needs: trigger
     uses: ./.github/workflows/linters.yml
 
+  coverage:
+    needs: trigger
+    runs-on: ubuntu-24.04
+    steps:
+    - uses: actions/checkout@v4
+    - name: Install Rust
+      run: rustup update stable
+    - name: Install cargo-llvm-cov
+      uses: taiki-e/install-action@cargo-llvm-cov
+    - name: Generate code coverage
+      run: |
+        sudo apt-get install -y libpcap-dev
+        cargo xtask code-coverage --context ci
+    - name: Upload coverage to Codecov
+      uses: codecov/codecov-action@v5
+      with:
+        token: ${{ secrets.CODECOV_TOKEN }}
+        files: lcov.info
+        fail_ci_if_error: true
+
   tests:
     needs: trigger
     uses: ./.github/workflows/tests.yml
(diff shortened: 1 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -19,6 +19,27 @@ jobs:
     needs: trigger
     uses: ./.github/workflows/linters.yml
 
+  coverage:
```
tomaszklak, line 22: would it be possible to use xtask so that it's also easy to generate coverage report locally?

.github/workflows/ci.yml
```diff
@@ -19,6 +19,27 @@ jobs:
...
+    steps:
+    - uses: actions/checkout@v4
+    - name: Install Rust
+      run: rustup update stable
+    - name: Install cargo-llvm-cov
+      uses: taiki-e/install-action@cargo-llvm-cov
+    - name: Generate code coverage
+      run: |
+        sudo apt-get install -y libpcap-dev
+        cargo llvm-cov clean --workspace
```
tomaszklak, line 34: We are running on CI, do we expect that there is anything to clean on a fresh container in which this job is running?

.github/workflows/ci.yml
tomaszklak: What I had in mind is that `cargo llvm-cov --all-features --workspace --lcov --output-path lcov.info --ignore-filename-regex xray\|integration\|xtask\|main` is complex enough, so that it would be nice to the other users of NepTUN to wrap up somehow. Not to mention that this is not the command you want to run locally - you probably want to drop `--lcov` and replace it with `--html`. But I see that xtask is very verbose, so the overhead of adding wrapper for running coverage locally is high. If you don't want to do that or think that this is bad idea, feel free to ignore this comment thread.

Hasan6979: To running it locally you can just run "cargo llvm-cov" and thats it. It prints details of every file, and that should be good. Instead of xtask, I can add a just file for this. Wdyt?

.github/workflows/ci.yml
Hasan6979: Good point. It isnt needed.

LukasPukenis (dismissed):
+1

.github/workflows/ci.yml
mathiaspeters: I don't think xtask is too verbose for that, and I think it's better to reuse the task runner we already have instead of adding more of them

Hasan6979: Add a xtask for code coverage with options [ci, local]. Local outputs a file in the html format which can be viewed in your browser. CI outputs a lcov.info file which can be read by code-coverage. Have kept it extensible for anything else that might come up.

PR #39, opened 2025-06-16
Have separate download/upload tests

Hasan6979:
Performing neptun <-> neptun tests is not the best way to benchmark performance because it hides where the bottlenecks could be (download or upload).

Instead it is better to isolate the upload and download case in its own tests with the other side as linux native wireguard.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -47,6 +47,9 @@ jobs:
 
   performance-tests:
     runs-on: ubuntu-24.04
+    strategy:
+      matrix:
+        test_type: [ upload, download ]
     steps:
     - uses: actions/checkout@a5ac7e51b41094c92402da3b24376905380afc29 # v4.1.6
       with:
@@ -55,7 +58,7 @@ jobs:
         sudo apt update
         sudo apt install -y libpcap-dev
     - run: docker build -t neptun-runner:0.0.1 .
-    - run: cargo xtask perf --base main
+    - run: cargo xtask perf --base main --test-type ${{ matrix.test_type }}
 
   xray-tests:
     runs-on: ubuntu-24.04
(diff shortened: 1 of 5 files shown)


.github/workflows/tests.yml
```diff
@@ -45,7 +45,21 @@ jobs:
     - uses: actions-rs/toolchain@b2417cde72dcf67f306c0ae8e0828a81bf0b189f # v1.0.6
     - run: cargo bench -p neptun --no-run
 
-  performance-tests:
+  download-performance-tests:
```
LukasPukenis, line 48: If I understand it right `TEST_TYPE` controls if it's upload or download - could we base one entry totally on another one and just override the variable? Or maybe even introduce something like:
`performance-tests-base:
...
`

and then for both upload and download cases refer to those?
Not a requirement though, it's good as it is since its already split, just offering to DRY it a bit if possible :)

xtask/perf/left.sh
```diff
@@ -7,6 +7,14 @@
 # in any real deployments                            #
 ######################################################
 
+if [ "$TEST_TYPE" = "download" ]; then
+    ip link add dev wg1 type wireguard
+    ip link add dev wg2 type wireguard
```
LukasPukenis, line 12: Just curious - why do we need to wg interfaces?

LukasPukenis (commented):
Can we also edit `perf.rs` to accept the variable or at least list that it's available if we manually run it?

xtask/perf/left.sh
Hasan6979: In order to use the native linux wireguard.

LukasPukenis: I mean one side is gonna be NepTUN while the other is gonna be native linux so I would expect that at most we need 1 native interface..

Hasan6979:
> Can we also edit `perf.rs` to accept the variable or at least list that it's available if we manually run it?

@LukasPukenis  Added `test_type` as an Optional argument in perf.rs.

LukasPukenis (approved):
+2

PR #40, opened 2025-06-16
Remove option + use Vec instead

Hasan6979:
Refactor of upload optimizations for neptun.

* Rename a few variables
* Remove Option<> from the EncryptionTaskData to ensure we don't end up sending None elements

### DIFF
--- a/neptun/src/device/mod.rs
+++ b/neptun/src/device/mod.rs
@@ -57,9 +57,10 @@ use thiserror::Error;
 
 const HANDSHAKE_RATE_LIMIT: u64 = 100; // The number of handshakes per second we can tolerate before using cookies
 
-const MAX_UDP_SIZE: usize = (1 << 16) - 1;
+// Max packet size of 1550 because packets are limited by the MTU sizes
+// used in wild networks.
+const MAX_PKT_SIZE: usize = 1550;
 const MAX_ITR: usize = 100;
-const PKT_SIZE: usize = 1600;
 const CHANNEL_SIZE: usize = 500;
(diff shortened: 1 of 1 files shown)


neptun/src/device/mod.rs
```diff
@@ -1149,9 +1137,9 @@ impl Device {
 
                 let peers = &d.peers_by_ip;
                 loop {
-                    let mut element = EncryptionTaskData::default();
+                    let mut buffer = [0u8; MAX_PKT_SIZE];
```
tomasz-grz, line 1140: I wonder, if rust compiler optimizations are smart enough to avoid repeated stack allocations in this loop?

Still looks better than creating a `EncryptionTaskData::default()` every iteration with 65535 byte buffer

neptun/src/device/mod.rs
```diff
@@ -57,9 +57,8 @@ use thiserror::Error;
 
 const HANDSHAKE_RATE_LIMIT: u64 = 100; // The number of handshakes per second we can tolerate before using cookies
 
-const MAX_UDP_SIZE: usize = (1 << 16) - 1;
+const MAX_PKT_SIZE: usize = 1600;
```
tomasz-grz, line 60: From `65535` to `1500`?
Can you please add a comment regarding the purpose of this change?

tomasz-grz (commented):
Just some minor questions otherwise looks alright +1

gytsto (commented):
+1

tomasz-grz (approved):
+1

PR #44, opened 2025-07-10
panic on unwrap

Hasan6979:
Android team faced a problem of panics at unwrap. It is a big difficult to get to the root cause of this, hence this issue is being hotfixed.

### DIFF
--- a/neptun/src/device/peer.rs
+++ b/neptun/src/device/peer.rs
@@ -94,7 +94,9 @@ impl Peer {
     pub fn shutdown_endpoint(&self) {
         if let Some(conn) = self.endpoint.write().conn.take() {
             tracing::info!("Disconnecting from endpoint");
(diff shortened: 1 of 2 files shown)


tomasz-grz (dismissed):
+2

neptun/src/device/peer.rs
```diff
@@ -94,7 +94,9 @@ impl Peer {
     pub fn shutdown_endpoint(&self) {
         if let Some(conn) = self.endpoint.write().conn.take() {
             tracing::info!("Disconnecting from endpoint");
-            conn.shutdown(Shutdown::Both).unwrap();
+            if let Err(e) = conn.shutdown(Shutdown::Both) {
+                tracing::error!("Error in conn shutdown. Exiting {}", e);
```
mathiaspeters, line 98: It won't be exiting though, it will just ignore the error

PR #48, opened 2025-07-17
[LLT-6456] handle unwraps

tomasz-grz:
## Problem

Code-base is using a lot of `unwrap()` causing a lot of panics in certain cases.

## Solution

Implement better error handling instead.

### DIFF
--- a/neptun/src/device/mod.rs
+++ b/neptun/src/device/mod.rs
@@ -576,33 +576,23 @@ impl Device {
         preshared_key: Option<[u8; 32]>,
     ) -> Result<Arc<Peer>, Error> {
         let next_index = self.next_index();
-        let device_key_pair = match self.key_pair.as_ref() {
-            Some(keypair) => keypair,
-            None => {
-                tracing::error!("No device keypair specified for a peer");
-                return Err(Error::InternalError(
-                    "No device keypair specified for a peer".to_owned(),
-                ));
-            }
-        };
+        let device_key_pair = self.key_pair.as_ref().ok_or_else(|| {
+            tracing::error!("No device keypair specified for a peer");
+            Error::InternalError("No device keypair specified for a peer".to_owned())
+        })?;
 
-        let tunn = match Tunn::new(
+        let tunn = Tunn::new(
             device_key_pair.0.clone(),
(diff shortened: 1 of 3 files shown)


Jauler (dismissed):
+1.0

But we may optionally consider adding some sort of string at startup in logs, like this: https://github.com/NordSecurity/NepTUN/commit/e1620ebb102dad16dfbb81d31ecef67afe3bc549

gytsto (commented):
Is there a linter job for unwraps enabled? If yes, then +1, otherwise would like to see that going :smile:

Jauler:
> Is there a linter job for unwraps enabled? If yes, then +1, otherwise would like to see that going 😄

Currently there is no nagger integration for NepTUN, hence we have created LLT-6457 as we would like to avoid blocking fixed for neptun panics in prod by linter work.

neptun/src/device/peer.rs
```diff
@@ -122,9 +124,14 @@ impl Peer {
             return Err(Error::Connect("Connected".to_owned()));
         }
 
-        let addr = endpoint
-            .addr
-            .expect("Attempt to connect to undefined endpoint");
+        let addr = if let Some(addr) = endpoint.addr {
```
mathiaspeters, line 127: This should be a match

tomasz-grz: Refactored

neptun/src/device/mod.rs
```diff
@@ -944,8 +967,9 @@ impl Device {
                     unsafe { &mut *(&mut t.src_buf[..] as *mut [u8] as *mut [MaybeUninit<u8>]) };
                 while let Ok((packet_len, addr)) = udp.recv_from(src_buf) {
                     let packet = &t.src_buf[..packet_len];
+
                     // The rate limiter initially checks mac1 and mac2, and optionally asks to send a cookie
-                    let parsed_packet =
+                    let parsed_packet = if let Some(rate_limiter) = rate_limiter {
```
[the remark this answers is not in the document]
tomasz-grz: Refactored

mathiaspeters (approved):
+1

PR #49, opened 2025-07-18
feat: log git sha on startup

tomasz-grz:
## Problem

We have no way of telling which version of NepTUN is running when debugging.

## Solution

Log the git sha during NepTUN startup.


mathiaspeters:
+1

LukasPukenis (approved):
It is a bit unexpected to have such a dependency diff size due to such a thing. In build.rs we could issue `git log -1` to reveals the last commit and then extract the hash with no dependency update, I wonder if this isn't more suitable :thinking: 

However the PR solves the problem :) +1

tomasz-grz:
@LukasPukenis very good point, I assumed the lock diff was because of the `criterion` version bump, because it was conflicting with some requirements of `vergen-git2` but I think your suggestion should be much more straight forward.

PR #50, opened 2025-07-21
Update version to v2.2.1

tomasz-grz:
## Problem

Package version in `Cargo.toml` is not corresponding to the released version numbers.

## Solution

Update package version in `Cargo.toml`

### DIFF
--- a/xray/Cargo.toml
+++ b/xray/Cargo.toml
@@ -20,6 +20,5 @@ tokio = { version = "1.41", features = ["macros", "rt-multi-thread", "time", "ne
 x25519-dalek = "2.0"
 
 [dependencies.neptun]
-version = "0.6.0"
 path = "../neptun"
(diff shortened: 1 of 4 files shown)


xray/Cargo.toml
```diff
@@ -20,6 +20,5 @@ tokio = { version = "1.41", features = ["macros", "rt-multi-thread", "time", "ne
 x25519-dalek = "2.0"
 
 [dependencies.neptun]
-version = "0.6.0"
```
Hasan6979, line 23: It might be obvious, but asking for my knowledge. Why is the version removed from here instead of updated to v2.2.1 ?

tomasz-grz: In my understanding it would be locked to `v2.2.1`, if we would like to run the tests with latest version we would have to keep updating it in multiple places.

PR #51, opened 2025-08-05
Add support for 32bit targets

LukasPukenis:
**Problem**
For OpenWRT we need to support MT7621AT chipset(32bit) which is powering few popular routers.  Though we could potentially omit NepTUN from teliod for OpenWRT altogether and leave just native linux module support, that would require much more rework. This fix seems to be pretty small and resolves the issue quickly - using the `ramips/mt7621` OpenWRT SDK, teliod compiles with this fix.

**Fix**
32bit targets do not have AtomicU64 hardware support but instead only have AtomicU32. As a solution to enable 32bit targets "AtomicType" is introduced which depends on the target.

### DIFF
--- a/neptun/src/noise/rate_limiter.rs
+++ b/neptun/src/noise/rate_limiter.rs
@@ -8,7 +8,6 @@ use crate::noise::{HandshakeInit, HandshakeResponse, Packet, Tunn, TunnResult, W
 #[cfg(feature = "mock-instant")]
 use mock_instant::Instant;
 use std::net::IpAddr;
-use std::sync::atomic::{AtomicU64, Ordering};
 
 #[cfg(not(feature = "mock-instant"))]
 use crate::sleepyinstant::Instant;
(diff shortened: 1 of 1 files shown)


Hasan6979 (commented):
+1

neptun/src/noise/rate_limiter.rs
```diff
@@ -43,13 +49,13 @@ pub struct RateLimiter {
     /// The key we use to derive the cookie
     secret_key: [u8; 16],
     start_time: Instant,
-    /// A single 64 bit counter (should suffice for many years)
-    nonce_ctr: AtomicU64,
+    /// A single 32/64 bit counter
+    nonce_ctr: AtomicType,
```
tomaszklak, line 53: I think that this change might significantly reduce the security of our implementation. Either we should get a green light for this change from the security team, or use Mutex on 32bit architectures.

tomaszklak (changes requested):

neptun/src/noise/rate_limiter.rs
tomaszklak: Jugging by the name of the field, it should never have the same value twice. Using u64 and incrementation is fine, since you will not be able to reach the same value this way in your lifetime. With u32 its trivial.

LukasPukenis: @tomaszklak thanks for a good point. I've replaced it into a Mutex after checking that the values are only used in handshake phase(init | response).

neptun/src/noise/rate_limiter.rs
```diff
@@ -44,12 +43,14 @@ pub struct RateLimiter {
     secret_key: [u8; 16],
     start_time: Instant,
     /// A single 64 bit counter (should suffice for many years)
-    nonce_ctr: AtomicU64,
+    /// AtomicU64 is not available for 32bit targets. Ensure not to use in hot-paths
+    nonce_ctr: Mutex<u64>,
```
tomaszklak, line 47: The comment above this structure explicitly talks about not using locking. I think that's because it's written from a pov of previous maintainers of this codebase, who primarily were concerned with server side usecase. In our, client side usecase the performance cost of locking is not an issue (if this is only used for handshakes).

Eitherway, please also update the comment so that it's not contradicting the structure definition.

LukasPukenis: Agreed on the comment, rewrote it to reflect better what is happening in the code.

PR #57, opened 2025-09-15
Use u64 instead of usize

Hasan6979:
Use u64 instead of usize in rx and tx bytes as the value overflows just after 4GB of data and incorrectly reports state which causing the connection to go from connected -> connecting


matislovas:
+1

stalowyjez (commented):
Hmm, do we run that on 32 bit machines?

Hasan6979:
@stalowyjez  Yes some Android TVs have 32bit processors. Which is from where the issue originated from 😉

stalowyjez:
Oki, +1.0

stalowyjez (approved):

PR #59, opened 2025-09-17
Add 32bit test in neptun

Hasan6979:
Add test to check rx and tx bytes don't overflow in 32bit systems

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -23,6 +23,19 @@ jobs:
       run: cargo +stable install --git https://github.com/taiki-e/cargo-hack.git cargo-hack --rev c0b517b9eefa27cdaf27cca5f1b186c00ef1af47 --locked
     - run: cargo hack test --each-feature ${{ matrix.packages }}
 
+  tests-32bit:
+    runs-on: ubuntu-24.04
+    steps:
+    - uses: actions/checkout@a5ac7e51b41094c92402da3b24376905380afc29 # v4.1.6
+    - uses: actions-rs/toolchain@b2417cde72dcf67f306c0ae8e0828a81bf0b189f # v1.0.6
+    - run: sudo dpkg --add-architecture i386 && sudo apt-get update
+    - run: sudo apt-get install -y gcc-multilib libc6:i386 libpcap-dev:i386
+    - name: Add 32bit target
+      run: rustup target add i686-unknown-linux-gnu
+    - name: Install hack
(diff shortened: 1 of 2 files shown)


neptun/src/noise/mod.rs
```diff
@@ -1082,4 +1091,35 @@ mod tests {
...
+            let data = if let TunnResult::WriteToNetwork(sent) = data {
+                sent
+            } else {
+                unreachable!();
+            };
+
+            their_tun.decapsulate(None, data, &mut their_dst);
+        }
+
+        let (_, _, rx, _, _) = their_tun.stats();
```
stalowyjez, line 1115: Why won't you merge these two let statements? 😅

neptun/src/noise/mod.rs
```diff
@@ -1082,4 +1091,35 @@ mod tests {
...
+    fn test_long_running_streams() {
+        let (mut my_tun, mut their_tun) = create_two_tuns_and_handshake();
+        let mut my_dst = [0u8; 1500];
+        let mut their_dst = [0u8; 1500];
+
+        let sent_packet_buf = create_large_ipv4_udp_packet();
+
+        for _ in 0..4_000_000 {
+            let data = my_tun.encapsulate(&sent_packet_buf, &mut my_dst);
+            assert!(matches!(data, TunnResult::WriteToNetwork(_)));
```
stalowyjez, line 1105: And this assert looks like it could be instead of `unreachable`

.github/workflows/tests.yml
```diff
@@ -23,6 +23,19 @@ jobs:
...
+    steps:
+    - uses: actions/checkout@a5ac7e51b41094c92402da3b24376905380afc29 # v4.1.6
+    - uses: actions-rs/toolchain@b2417cde72dcf67f306c0ae8e0828a81bf0b189f # v1.0.6
+    - run: sudo dpkg --add-architecture i386 && sudo apt-get update
+    - run: sudo apt-get install -y gcc-multilib libc6:i386 libpcap-dev:i386
+    - name: Add 32bit target
+      run: rustup target add i686-unknown-linux-gnu
+    - name: Install hack
+      run: cargo +stable install --git https://github.com/taiki-e/cargo-hack.git cargo-hack --rev c0b517b9eefa27cdaf27cca5f1b186c00ef1af47 --locked
+    - run: cargo hack test --each-feature ${{ matrix.packages }} --target i686-unknown-linux-gnu
```
stalowyjez, line 37: Maybe we could have some container with this stuff installed instead of doing it every time from scratch?

stalowyjez (commented):
Look good, just a few things to consider

neptun/src/noise/mod.rs
stalowyjez: Oh, nvm

Issue #63, opened 2026-01-21
Consider git checking optional in build phase.


ilsubyeega:
https://github.com/NordSecurity/NepTUN/blob/b888a5786101f3996afb2409212a0be176cdfc92/neptun/build.rs#L6-L13

Hello, i'm packaging nordvpn-linux -> libtello and the `build.rs` seems to call git but no fallback there

ilsubyeega:
> Hey! Can you specify, what error are you getting here?

Hi, mostly building crates when offline-sandboxed(i believe its via `cargo vendor` iirc) do not have `.git` directory so the git will be failed

PR #66, opened 2026-03-18
Enable building without git history

matislovas:
Some build systems [strips](https://github.com/NordSecurity/NepTUN/issues/63) of git history. Our `build.rs` script depends on git. So we need a fallback in case of no git history.


tomasz-grz (commented):
+1

Jauler (approved):
+1.0

PR #67, opened 2026-03-30
LLT-7073 Add Apple platform specific packet processing

djkarwowski:
### Problem
- Current NepTUN implementation shows degraded performance on Apple platforms
- Root cause
  - upload: multiple workers spawned without QoS hints land on mixed P/E cores, causing concurrent batch processing which results in out-of-order packet delivery and TCP congestion control triggering repeated CWND halving
  - download: per-packet heap allocations in the iface handler continuously evict the UDP handler's working set from L1D cache, causing sustained cache-cold decryption throughput reduction

### Solution
- Restore inline, single-threaded iface handler on Apple platforms using thread-local buffers to avoid cache eviction pressure on UDP handler

### DIFF
--- a/neptun/src/device/mod.rs
+++ b/neptun/src/device/mod.rs
@@ -222,6 +222,39 @@ struct TunnelWorkerData {
     buf_len: usize,
 }
 
+enum IfaceReadResult<'a> {
+    Packet { payload: &'a [u8], peer: Arc<Peer> },
+    Exhausted,
+    Fatal,
+    Skip,
+}
+
+#[cfg(not(any(target_os = "macos", target_os = "ios", target_os = "tvos")))]
+enum BatchResult {
+    Continue,
+    Exhausted,
+    Fatal,
+}
+
+struct UnexpectedEncapsulationResult;
+
+struct CheckedMtu(usize);
+
+impl CheckedMtu {
+    fn new(mtu: usize) -> Option<Self> {
+        if mtu + WG_HEADER_OFFSET > MAX_PKT_SIZE {
+            tracing::error!("Insufficient packet buffer size");
+            None
+        } else {
+            Some(Self(mtu))
+        }
+    }
+
+    fn get(&self) -> usize {
+        self.0
+    }
+}
+
 #[cfg(not(any(target_os = "macos", target_os = "ios", target_os = "tvos")))]
 type EventLoopThreads = Result<(Vec<JoinHandle<()>>, Arc<Lock<Vec<Arc<TunSocket>>>>), Error>;
 #[cfg(any(target_os = "macos", target_os = "ios", target_os = "tvos"))]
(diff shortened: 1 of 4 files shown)


neptun/src/device/mod.rs
```diff
@@ -48,6 +48,7 @@ use std::os::fd::{AsFd, BorrowedFd, RawFd};
 use std::os::unix::io::AsRawFd;
 use std::sync::atomic::{AtomicUsize, Ordering};
 use std::sync::Arc;
+#[cfg(not(any(target_os = "macos", target_os = "ios", target_os = "tvos")))]
```
matislovas: You can combine std::thread and std::thread::JOinHandle to be under one cfg flag.

djkarwowski: Thanks for pointing this out.

neptun/src/device/peer.rs
```diff
@@ -114,9 +118,14 @@ impl Peer {
         endpoint.addr = Some(addr);
     }
 
+    /// On Apple platforms, setting `skt_buffer_size` won't have any effect whatsoever.
```
Jauler: I believe the effect will be that the buffer is configured as per `setsockopt()`. But manual configuration of socket buffer sizes will likely be limited to significantly smaller sizes compared to kernel autotuning. Hence the end-result is likely to be worse, especially on long-fat-networks.

Hence maybe it is worth clarifying this in the comment, just to avoid possible confusion.

neptun/src/device/mod.rs
```diff
@@ -1213,6 +1247,131 @@ impl Device {
         Ok(())
     }
 
+    #[cfg(any(target_os = "macos", target_os = "ios", target_os = "tvos"))]
+    fn register_read_iface_handler(&self, iface: Arc<TunSocket>) -> Result<(), Error> {
```
Jauler, line 1191: I am thinking, that it would be nice, if reasonable - to somehow avoid the duplication of full `register_read_iface_handler()` function.

To my understanding, and correct me if I am wrong - the main difference between apple version and android version of this function is that one is reading some amount of packets (or until EWOULDBLOCK is return) from the tun `fd` and sending that batch over channel for another thread for encryption. While apple version does encryption in-place. 

The rest of the function is pretty much the same. 

Hence I am thinking maybe we could refactor something one the lines of:
- create `handle_packet_queued()` and `handle_packet_in_place()`
- call them from within the `register_read_iface_handler()` depending on the platform

Hopefully this way we would export platform-specific functionality into smaller functions and do conditional compilation there, reducing the duplication.

djkarwowski: Thank you for your valuable observation and hints. This is indeed the key difference and I fully agree with the suggestion. Hopefully, the refactored solution addresses this issue and expresses the intent behind the code more clearly.

neptun/src/device/peer.rs
djkarwowski: Please let me know if the updated comment expresses the intent more properly now.

gytsto (commented):
Would love to see CI all green first before approving

neptun/src/device/mod.rs
```diff
@@ -778,6 +832,18 @@ impl Device {
 
         let rx_clone = self.socket_to_tunnel_rx.clone();
         let fw_callback = self.config.firewall_process_inbound_callback.clone();
+
+        #[cfg(any(target_os = "macos", target_os = "ios", target_os = "tvos"))]
+        {
+            let queue = dispatch::Queue::global(dispatch::QueuePriority::High);
```
matislovas: Shouldn't we store these queues somewhere and on whole thing shuting down, call leave() or "await" ..?

neptun/src/device/mod.rs
```diff
@@ -1223,76 +1289,57 @@ impl Device {
...
+
+                    loop {
+                        let (batched_pkts, result) =
+                            handle_packet_queued(&iface, &mtu, peers, max_batched_pkts);
+
+                        match result {
+                            BatchResult::Fatal => return Action::Exit,
+                            BatchResult::Continue | BatchResult::Exhausted => {
+                                if let Err(e) = d.tunnel_to_socket_tx.send(batched_pkts) {
+                                    tracing::warn!(
```
matislovas: On Apple path, you have:

```
  if encapsulate_and_send(&peer, &mut thread_data.dst_buf[..], len, udp4, udp6)
      .is_err()
  {
      return Action::Exit;
  }
```

But on rest of the platforms there's only and log entry on failed send, why the difference?

djkarwowski: Good catch, I have aligned the approach between platforms with regard to this point.

neptun/src/device/mod.rs
djkarwowski: This worker is only used for connected sockets. Technically it means that `use_connected_socket` on the `Device`'s config should be set to `true` in order to send packets for processing to the worker thread defined here. However, this contradicts the inline processing approach we take for Apple platforms. Indeed, in `libtelio` we set the `use_connected_socket` to false for iOS, macOS and tvOS ([see here](https://github.com/NordSecurity/libtelio/blob/c70ad1f636b227379088266f3bc489eb2f4bdc03/crates/telio-wg/src/adapter/neptun.rs#L62)), so practically the worker thread we create here stays idle on these platforms anyway. Hence, I decided to remove this part at all.

neptun/src/device/mod.rs
```diff
@@ -138,6 +141,8 @@ pub struct DeviceHandle {
 #[derive(Clone)]
 pub struct DeviceConfig {
     pub n_threads: usize,
+    /// On Apple platform, packets are always handled through an anonymous UDP socket.
```
Jauler: This is a small nitpick, but IMHO would be nice to change "anonymous" to "unconnected" - as it is a bit confusing what is meant by anonymous :). 

```suggestion
    /// On Apple platform, packets are always handled through an unconnected UDP socket.
```

neptun/src/device/mod.rs
```diff
@@ -739,6 +732,7 @@ impl Device {
         udp_sock6.set_nonblocking(true)?;
         self.config.protect.make_external(udp_sock6.as_raw_fd());
 
+        #[cfg(not(any(target_os = "macos", target_os = "ios", target_os = "tvos")))]
```
Jauler, line 735: While it is true, that manually adjusting socket buffer sizes on apple is significantly more limited compared to kernel autotuning (autotuning may increase sizes to significantly more than setsockopt), allowing for larger [BDP](https://en.wikipedia.org/wiki/Bandwidth-delay_product)s to be supported - but I am not sure if conditionally compiling-out the configuration values (which are optional) is useful TBH. 

What do you think, does it make sense to remove them completely? Or just leave them unconfigured, but avoid the conditional compilation directives?

neptun/src/device/mod.rs
```diff
@@ -1303,138 +1227,126 @@ impl Device {
...
+        TunnResult::Done => {}
+        TunnResult::Err(e) => {
+            tracing::error!(message = "Encapsulate error",
+                error = ?e,
+                public_key = peer.public_key.1);
+        }
+        TunnResult::WriteToNetwork(packet) => {
+            let endpoint = peer.endpoint();
+            if let Some(conn) = endpoint.conn.as_ref() {
+                if let Err(err) = conn.send(packet) {
```
Jauler: I believe our `endpoint.conn` is actually set to nonblocking mode, which means, that in this conn.send() returns `EWOULDBLOCK` (or equivalent error) - we would drop the connected socket. 

This is not really correct, even though this _was_ the case on main as well. Maybe it would make sense to fix it. I am thinking about two possible approaches:
- One is to drop the packet in case of SNDBUF being full
- The other is to use blocking mode for the socket

 In practice the difference between the two seems to be rather minimal. In one case we are dropping packets at `conn.send()` point, the other is at `/dev/net/tun` or `utun` queue. But IMHO dropping the connected socket (and reconnecting it on further transmissions) is probably not what is intended :).

Jauler (dismissed):
There are a few minor comments, but overall - changes looks nice. Great work!

neptun/src/device/mod.rs
djkarwowski: The idea was that entirely removing `modify_skt_buffer_size()` function from compilation will put us on the safe side, preventing us from attempts to manually adjust socket buffer sizes for Apple platforms.

At the same time I agree with your point, especially that conditionally removing `modify_skt_buffer_size()` function from compilation implies even more cfg gating in several other places (this one included).

However, I still think it may be good to be explicit about that. Maybe instead of relying on conditional compilation, we could simply make it a no-op on Apple?

```
fn modify_skt_buffer_size(socket: BorrowedFd<'_>, buffer_size: usize) {
    // For Apple, rely on the kernel autotuning of the socket buffer size
    if cfg!(any(
        target_os = "macos",
        target_os = "ios",
        target_os = "tvos"
    )) {
        return;
    }

    set_sock_opt(socket, NixSocket::sockopt::RcvBuf, buffer_size, "RcvBuf");
    set_sock_opt(socket, NixSocket::sockopt::SndBuf, buffer_size, "SndBuf");
}
```

Do you think this makes sense?

Jauler: :thinking: 

Maybe I am overthinking this. Lets just leave it as is. Its safe, and kind of clearly indicates intention of not doing skt buffer changes on apple.

neptun/src/device/mod.rs
djkarwowski: Updated

neptun/src/device/mod.rs
djkarwowski: I went with the first approach of dropping a packet on `EWOULDBLOCK` but kept the original behavior of dropping connection on other errors. Please let me know if this is expected.

djkarwowski: Also, `register_read_conn_skt_handler()` uses a connected socket, and only [drops packet on send error](https://github.com/NordSecurity/NepTUN/blob/688f8c0be836f5ad3bc9f59364c8ed1af6358159/neptun/src/device/mod.rs#L1140). Should we align the behavior in these two cases (and possibly the logging level as well)?

Jauler:
+1.0

Jauler (approved):
+1.0

gytsto:
+1

PR #69, opened 2026-05-20
LLT-7323: base64 validation

tomasz-grz:
## Problem
The KeyBytes parser in `serialization.rs` had several security issues and was missing better base64 validation.

## Solution
- Added `base64` validation, updated dependency and removed unnecessary heap allocation.
- Added `zeroize` to ensure key memory is zeroed when dropped.
- Made the inner field private and added a `get_bytes()` function. This ensures `Drop` always runs and that `KeyBytes` can not be constructed directly, bypassing validation.

### DIFF
--- a/neptun/src/device/api.rs
+++ b/neptun/src/device/api.rs
@@ -248,7 +248,7 @@ fn api_set<R: Read>(reader: &mut BufReader<R>, d: &mut LockReadGuard<Device>) ->
                     match key {
                         "private_key" => match val.parse::<KeyBytes>() {
                             Ok(key_bytes) => {
-                                device.set_key(x25519::StaticSecret::from(key_bytes.0))
+                                device.set_key(x25519::StaticSecret::from(*key_bytes.get_bytes()))
                             }
                             Err(_) => return EINVAL,
                         },
@@ -285,7 +285,7 @@ fn api_set<R: Read>(reader: &mut BufReader<R>, d: &mut LockReadGuard<Device>) ->
                                 return api_set_peer(
                                     reader,
                                     device,
-                                    x25519::PublicKey::from(key_bytes.0),
+                                    x25519::PublicKey::from(*key_bytes.get_bytes()),
                                 )
                             }
                             Err(_) => return EINVAL,
@@ -351,7 +351,7 @@ fn api_set_peer<R: Read>(
                     Err(_) => return EINVAL,
                 },
                 "preshared_key" => match val.parse::<KeyBytes>() {
-                    Ok(key_bytes) => preshared_key = Some(key_bytes.0),
+                    Ok(key_bytes) => preshared_key = Some(*key_bytes.get_bytes()),
                     Err(_) => return EINVAL,
                 },
                 "endpoint" => match val.parse::<SocketAddr>() {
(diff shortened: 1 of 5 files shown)


gytsto (commented):
+1

neptun/src/device/api.rs
```diff
@@ -351,7 +351,7 @@ fn api_set_peer<R: Read>(
                     Err(_) => return EINVAL,
                 },
                 "preshared_key" => match val.parse::<KeyBytes>() {
-                    Ok(key_bytes) => preshared_key = Some(key_bytes.0),
+                    Ok(key_bytes) => preshared_key = Some(*key_bytes.get_bytes()),
```
lcruz99, line 354: This self-defeats the purpose of zeroizing `key_bytes`. When you call `key_bytes.get_bytes()` it'll get a borrow of the inner array `&[u8; 32]` but when you dereference it it'll just copy it to `preshared_key` which lives in the stack frame of `api_set_peer`.

To fix this you can either wrap the local `preshared_key` also in `Zeroizing`:

```
  let mut preshared_key: Option<Zeroizing<[u8; 32]>> = None;
  // ...
  "preshared_key" => match val.parse::<KeyBytes>() {
      Ok(key_bytes) => preshared_key = Some(Zeroizing::new(*key_bytes.get_bytes())),
      Err(_) => return EINVAL,
  },
```

or add a new method to properly move the inner array as a zeroized value.

``` impl KeyBytes {
      pub(crate) fn into_inner(self) -> Zeroizing<[u8; 32]> {
          Zeroizing::new(self.0)
      }
  }
```

neptun/src/serialization.rs
```diff
@@ -20,19 +35,163 @@ impl std::str::FromStr for KeyBytes {
...
             }
-            43 | 44 => {
-                // Try to parse as base64
-                if let Ok(decoded_key) = base64::decode(s) {
-                    if decoded_key.len() == internal.len() {
-                        internal[..].copy_from_slice(&decoded_key);
-                    } else {
-                        return Err("Illegal character in key");
-                    }
+            43 => {
```
Jauler: TBH I am thinking, that maybe it is worth dropping support for unpadded base64? It is not supported by other WG implementations anyways 🤔🤔🤔

tomasz-grz: Good idea!

neptun/src/device/api.rs
tomasz-grz: Like you said, it lives in stack frame of `api_set_peer`, instead of lingering on the heap.
Which significantly reduces it's footprint compared to before, so I would disagree that it defeats the purpose entirely.

`preshared_key` defined as `Option<[u8; 32]>` (compared to `pub_key` and `private_key` that use `x25519::StaticSecret`)  so zeroizing it here would not be a significant improvement, as it would live on through out the rest of the lifetime of [Peer](https://github.com/NordSecurity/NepTUN/blob/1905cd621c6b5247852d48c998ae2066d7601a58/neptun/src/device/peer.rs#L32).

I don't know how sensitive is `preshared_key`, but if it was, I would expect it to wrap `x25519` just like pub and private key does 🤔

tomasz-grz: Propagating `Zeroizing` through the entire chain would be the ideal fix, but it's a large refactor that I would rather not do in the scope of this task.

neptun/src/serialization.rs
tomasz-grz: Updated

neptun/src/device/api.rs
tomasz-grz: Registered ticket for it. LLT-7354

lcruz99 (approved):
+1

PR #70, opened 2026-05-24
LLT-7202: Include integration tests in code coverage

djkarwowski:
### Problem
- The coverage job in the CI pipeline doesn't run integration tests for the `device` module. As a result, the report does not fully reflect the actual code coverage ensured by the test suite.
- Running `device` integration tests and generating code coverage report locally is not supported by the dedicated `xtask` at the moment.
- Integration tests are part of the same module as the tested code, so they appear as unit tests from Cargo perspective.
- Integration tests for the `device` module require Docker and `sudo` privileges, so it is desired to exclude them by default. Currently this is done by using `#[ignore]` attribute, which may be misinterpreted. Also, these tests are excluded from running on macOS. While in the CI this is desirable (see [Limitations](#Limitations) below), they can be run successfully in local development environment.

### Solution
- Moved all integration tests to a dedicated `tests` directory, so that Cargo naturally discovers them as integration tests.
- Instead of relying on `#[ignore]` attribute, created a dedicated `docker-tests` feature to make it both easy for opting-in for `device` integration tests and skipping them if needed (_e.g._, for running `cargo hack test --each-feature`).
- Created a dedicated `cargo xtask test` command to simplify local run of full test suite and its subsets on different platforms.
- Updated `cargo xtask code-coverage` command to create a composable pipeline allowing for running code coverage with different set of features.
- Included `device` module integration tests for Linux in the CI coverage job.

### Limitations
- Integration tests for the `device` module are excluded from the macOS coverage report in the CI. The reason is that GitHub's Apple Silicon runners don't have Docker support.

### DIFF
--- a/xtask/src/codecoverage.rs
+++ b/xtask/src/codecoverage.rs
@@ -1,46 +1,111 @@
 use clap::{Parser, ValueEnum};
-use std::process::Command;
+use std::{
+    io,
+    process::{Command, ExitStatus},
+};
+
+use crate::test::get_platform_cfg;
 
 #[derive(Parser, Debug)]
 pub struct Cmd {
-    /// Coverage run context ("local", "ci")
-    #[arg(
-        long,
-        required = true,
-        help = "Context of test run (\"local\" or \"ci\"). Defaults to \"local\""
-    )]
-    context: RunContext,
+    /// Coverage run action (defaults to `local` if no action is specified)
+    action: Option<Action>,
 }
 
 #[derive(Debug, Clone, ValueEnum)]
-pub enum RunContext {
+pub enum Action {
+    /// Clean accumulated coverage data
+    Clean,
+    /// Run tests with `mock-instant` feature enabled and accumulate coverage data
+    MockInstant,
+    /// Run tests with `docker-tests` feature enabled and accumulate coverage data
+    /// (requires sudo privileges and Docker)
+    DockerTests,
+    /// Emit an lcov report from accumulated coverage data
+    ReportLcov,
+    /// Emit an HTML report from accumulated coverage data and open it in a browser
+    ReportHtml,
+    /// Run the full code coverage pipeline, store the report and open it in a browser
     Local,
-    Ci,
 }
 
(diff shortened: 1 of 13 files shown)


matislovas:
+1

Paintree:
It may not be worth splitting  "Refactor integration tests for NepTUN" into two parts if part one is a dozen line of code, and part two is hundreds

xtask/src/codecoverage.rs
```diff
@@ -1,46 +1,111 @@
...
 pub struct Cmd {
-    /// Coverage run context ("local", "ci")
-    #[arg(
-        long,
-        required = true,
-        help = "Context of test run (\"local\" or \"ci\"). Defaults to \"local\""
-    )]
-    context: RunContext,
+    /// Coverage run action (defaults to `local` if no action is specified)
+    context: Option<Action>,
```
Paintree: Maybe `context` is worth renaming to `action`?

Paintree (dismissed):
+1

djkarwowski:
@matislovas According to our discussion, I have restored the `coverage` job in the `ci.yml` as an umbrella job.

xtask/src/codecoverage.rs
djkarwowski: Definitely! Thanks!

djkarwowski:
> It may not be worth splitting "Refactor integration tests for NepTUN" into two parts if part one is a dozen line of code, and part two is hundreds

@Paintree The idea behind the split is as follows:
* Part 1 is primarily about moving files to a separate directory. The changes were kept to minimum, so that these files are actually recognized as moved rather than deleted/created.
* Part 2 introduces the actual changes in the integration test files. One of these changes is removing a redundant nesting, so most of these modified lines differ in whitespace only. If you look at the diff between commits for Part 1 and 2 locally in a tool like, _e.g._, VSCode, it is displayed in a much nicer way, showing only the actually meaningful diff.

Hopefully this makes some sense :)

Paintree (approved):

PR #71, opened 2026-05-29
Update firewall callback signature

jjanowsk:
New version of libfirewall uses mutable buffer in outbound callback

### DIFF
--- a/neptun/src/device/mod.rs
+++ b/neptun/src/device/mod.rs
@@ -151,7 +151,7 @@ pub struct DeviceConfig {
     pub firewall_process_inbound_callback:
         Option<Arc<dyn Fn(&[u8; 32], &mut [u8]) -> bool + Send + Sync>>,
     pub firewall_process_outbound_callback:
-        Option<Arc<dyn Fn(&[u8; 32], &[u8], &mut dyn std::io::Write) -> bool + Send + Sync>>,
+        Option<Arc<dyn Fn(&[u8; 32], &mut [u8], &mut dyn std::io::Write) -> bool + Send + Sync>>,
     pub skt_buffer_size: Option<usize>,
     pub inter_thread_channel_size: Option<usize>,
     pub max_inter_thread_batched_pkts: Option<usize>,
@@ -198,7 +198,10 @@ struct ThreadData {
 }
 
 enum IfaceReadResult<'a> {
-    Packet { payload: &'a [u8], peer: Arc<Peer> },
+    Packet {
+        payload: &'a mut [u8],
+        peer: Arc<Peer>,
+    },
     Exhausted,
     Fatal,
     Skip,
--- a/neptun/src/device/packet_workers.rs
+++ b/neptun/src/device/packet_workers.rs
@@ -210,7 +210,7 @@ fn write_to_socket_worker(
     udp4: Arc<socket2::Socket>,
     udp6: Arc<socket2::Socket>,
     firewall_process_outbound_callback: Option<
-        Arc<dyn Fn(&[u8; 32], &[u8], &mut dyn std::io::Write) -> bool + Send + Sync>,
+        Arc<dyn Fn(&[u8; 32], &mut [u8], &mut dyn std::io::Write) -> bool + Send + Sync>,
     >,
 ) {
     loop {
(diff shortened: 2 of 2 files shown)


neptun/src/device/mod.rs
```diff
@@ -1249,13 +1249,18 @@ fn process_iface_inline(
             IfaceReadResult::Fatal => return Action::Exit,
             IfaceReadResult::Skip => continue,
             IfaceReadResult::Packet { payload, peer } => {
+                let len = payload.len();
+
                 if let Some(callback) = &device.config.firewall_process_outbound_callback {
-                    if !callback(&peer.public_key.0, payload, &mut thread_data.iface.as_ref()) {
+                    if !callback(
+                        &peer.public_key.0,
+                        &mut thread_data.dst_buf[..],
```
sfraczek: Payload is a slice starting at `WG_HEADER_OFFSET`. Is it intentional that now you pass the whole buffer from index 0?

jjanowsk: I just also realized that :D Fixed already

neptun/src/device/mod.rs
```diff
@@ -1249,13 +1249,18 @@ fn process_iface_inline(
             IfaceReadResult::Fatal => return Action::Exit,
             IfaceReadResult::Skip => continue,
             IfaceReadResult::Packet { payload, peer } => {
+                let len = payload.len();
```
tomaszpatejko: It's difficult to notice a clear relation between `payload` and `thread_data.dst_buf` just by looking at this code block. So maybe it would be ok to use just `thread_data.dst_buf`?

sfraczek (commented):
+1

lcruz99 (dismissed):
+1

neptun/src/device/mod.rs
djkarwowski: I think it could be easier to make `payload` mutable in the `IfaceReadResult` definition:
```
enum IfaceReadResult<'a> {
    Packet { payload: &'a mut [u8], peer: Arc<Peer> },
    Exhausted,
    Fatal,
    Skip,
}
```

and keep passing the `payload`.

I see two benefits of this approach:
* `payload` semantically already refers to the packet data, making it clear the callback operates on the exact data that was read
* eliminates reconstructing the slice, avoiding an additional unchecked indexing operation

neptun/src/device/mod.rs
djkarwowski: Alternatively, I think we can still take advantage of the already existing `IfaceReadResult` type, as proposed [here](https://github.com/NordSecurity/NepTUN/pull/71#discussion_r3333510095).

tomaszpatejko: @djkarwowski yes, I think what you proposed makes more sense.

neptun/src/device/mod.rs
djkarwowski, line 1271: This is not related to your change - as a matter of fact, it is related to mine :) When looking through your PR I spotted something that could be tightened up, if that's OK for you to fix it here (otherwise please just resolve this comment and I'll raise a separate PR).

The comment I put on allowing the indexing slicing below does not reflect the full picture. It puts focus on `CheckedMtu` type as a guard ensuring the safe indexing slicing, but `buf` is accepted as `&mut [u8]`, so nothing in the type system actually prevents a caller from passing a buffer smaller than `WG_HEADER_OFFSET + mtu`.

We could harden this check by changing the `buf` signature to:
```
buf: &'a mut [u8; MAX_PKT_SIZE],
```

and also update the comment to something like this:
```
// buf is [u8; MAX_PKT_SIZE] and CheckedMtu guarantees mtu + WG_HEADER_OFFSET <= MAX_PKT_SIZE
```

djkarwowski (commented):
+1

neptun/src/device/mod.rs
jjanowsk: Sorry I'm on tight schedule trying to merge a feature that require changes in 6 repos so I won't do it for you this time

djkarwowski: No worries, sorry for asking! :)

djkarwowski (approved):

PR #73, opened 2026-07-06
Move from patched dispatch to dispatch2

stalowyjez:
It's already upgraded in Telio, so let's upgrade it in NepTUN too

### DIFF
--- a/neptun/src/device/mod.rs
+++ b/neptun/src/device/mod.rs
@@ -40,7 +40,9 @@ use rand_core::{OsRng, RngCore};
 use socket2::{Domain, Protocol, Type};
 use std::collections::HashMap;
 use std::io::{self, BufReader, BufWriter, Write};
-use std::mem::{swap, MaybeUninit};
+#[cfg(not(any(target_os = "macos", target_os = "ios", target_os = "tvos")))]
+use std::mem::swap;
+use std::mem::MaybeUninit;
 use std::net::{Ipv4Addr, Ipv6Addr, SocketAddr, SocketAddrV4, SocketAddrV6};
 use std::os::fd::RawFd;
 #[cfg(not(target_os = "windows"))]
@@ -49,6 +51,12 @@ use std::sync::atomic::{AtomicUsize, Ordering};
 use std::sync::Arc;
 use thiserror::Error;
 use tun::TunSocket;
+
+#[cfg(any(target_os = "macos", target_os = "ios", target_os = "tvos"))]
+use dispatch2::{
+    DispatchGroup, DispatchQueue, DispatchQueueGlobalPriority, DispatchRetained, DispatchTime,
+    GlobalQueueIdentifier,
+};
 
 #[cfg(not(any(target_os = "macos", target_os = "ios", target_os = "tvos")))]
 use {
@@ -132,7 +140,7 @@ pub struct DeviceHandle {
     #[cfg(not(any(target_os = "macos", target_os = "ios", target_os = "tvos")))]
     threads: Vec<thread::JoinHandle<()>>,
     #[cfg(any(target_os = "macos", target_os = "ios", target_os = "tvos"))]
-    threads: (dispatch::Group, Vec<dispatch::Queue>),
+    threads: DispatchRetained<DispatchGroup>,
     sockets_to_close: Arc<Lock<Vec<Arc<TunSocket>>>>,
 }
 
@@ -227,7 +235,7 @@ type EventLoopThreads = Result<(Vec<JoinHandle<()>>, Arc<Lock<Vec<Arc<TunSocket>
 #[cfg(any(target_os = "macos", target_os = "ios", target_os = "tvos"))]
 type EventLoopThreads = Result<
     (
-        (dispatch::Group, Vec<dispatch::Queue>),
+        DispatchRetained<DispatchGroup>,
         Arc<Lock<Vec<Arc<TunSocket>>>>,
     ),
     Error,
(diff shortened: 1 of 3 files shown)


neptun/src/device/mod.rs
```diff
@@ -132,7 +132,10 @@ pub struct DeviceHandle {
     #[cfg(not(any(target_os = "macos", target_os = "ios", target_os = "tvos")))]
     threads: Vec<thread::JoinHandle<()>>,
     #[cfg(any(target_os = "macos", target_os = "ios", target_os = "tvos"))]
-    threads: (dispatch::Group, Vec<dispatch::Queue>),
+    threads: (
+        dispatch2::DispatchRetained<dispatch2::DispatchGroup>,
+        Vec<dispatch2::DispatchRetained<dispatch2::DispatchQueue>>,
+    ),
```
lcruz99: maybe you can just create an alias out of this, here and on the following changes

neptun/src/device/mod.rs
```diff
@@ -1445,4 +1474,47 @@ mod tests {
...
+        let (threads, _sockets_to_close) =
+            DeviceHandle::start_event_loop_threads(N_THREADS, Arc::clone(&interface_lock)).unwrap();
+
+        // The migrated dispatch2 code allocates one global queue per thread.
+        assert_eq!(threads.1.len(), N_THREADS);
+
+        // Stop the loops and join the group. If the `enter()` guard were dropped
+        // early (the classic dispatch2 porting bug), the loops would never be
+        // members of the group and `wait()` would return before they exit — so
+        // this teardown also pins down the guard/wait semantics.
```
lcruz99: ` Stop the loops and join the group.` I think this sentence would suffice

lcruz99:
other than that, +1

neptun/src/device/mod.rs
```diff
@@ -635,11 +646,29 @@ impl Device {
...
     pub fn new_with_tun(tun: TunSocket, config: DeviceConfig) -> Result<Device, Error> {
-        let poll = EventPoll::<Handler>::new()?;
-
         // Create a tunnel device
         let iface = Arc::new(tun.set_non_blocking()?);
         let mtu = iface.mtu()?;
+        Self::new_with_iface(iface, mtu, config)
+    }
+
+    #[cfg(test)]
```
sfraczek: dead code on other platforms
```rust
#[cfg(all(test, any(target_os = "macos", target_os = "ios", target_os = "tvos")))]
```

neptun/src/device/mod.rs
```diff
@@ -1445,4 +1474,47 @@ mod tests {
...
+
+        // The migrated dispatch2 code allocates one global queue per thread.
+        assert_eq!(threads.1.len(), N_THREADS);
+
+        // Stop the loops and join the group. If the `enter()` guard were dropped
+        // early (the classic dispatch2 porting bug), the loops would never be
+        // members of the group and `wait()` would return before they exit — so
+        // this teardown also pins down the guard/wait semantics.
+        interface_lock.read().trigger_exit();
+        let _ = threads.0.wait(dispatch2::DispatchTime::FOREVER);
```
sfraczek: Please try to make this wait for few seconds and verify that it didn't timeout but succeeded. On error this could hang forever.
```rust
assert!(threads.0.wait(DispatchTime::try_from(Duration::from_secs(10)).unwrap()).is_ok(), "event loops did not exit");
```

neptun/src/device/mod.rs
```diff
@@ -1445,4 +1474,47 @@ mod tests {
...
+            max_inter_thread_batched_pkts: None,
+        };
+
+        let device = Device::new_with_mock_tun(tun, MOCK_MTU, config).unwrap();
+        let interface_lock = Arc::new(Lock::new(device));
+
+        let (threads, _sockets_to_close) =
+            DeviceHandle::start_event_loop_threads(N_THREADS, Arc::clone(&interface_lock)).unwrap();
+
+        // The migrated dispatch2 code allocates one global queue per thread.
```
sfraczek: It copies the pointer to the shared global process-wide queue. Consider fetching the queue once before the loop. Also consider storing only the group in threads instead of the tuple (group, queues) as nothing ever reads the queues because wait() only uses the group.

neptun/src/device/mod.rs
```diff
@@ -1445,4 +1474,47 @@ mod tests {
...
+        assert_eq!(threads.1.len(), N_THREADS);
```
sfraczek: this is always true and doesn't prove anything - loop always runs and pushes 4 times

neptun/src/device/mod.rs
```diff
@@ -273,9 +279,14 @@ impl DeviceHandle {
...
-                    let queue = dispatch::Queue::global(dispatch::QueuePriority::High);
+                    let queue = dispatch2::DispatchQueue::global_queue(
+                        dispatch2::GlobalQueueIdentifier::Priority(
+                            dispatch2::DispatchQueueGlobalPriority::High,
+                        ),
+                    );
                     queue.exec_async(move || {
-                        group_clone.enter();
+                        // Hold the group-membership guard for the whole event loop
+                        let _group_guard = group_clone.enter();
```
sfraczek: enter() still runs only when GCD starts the block, not at submit time. Between exec_async() returning and the blocks starting, the group count is 0, so wait() can return immediately.
dispatch2 has `dispatch_group_async` built in, which associates the block with the group atomically at submit time: `group.exec_async(&queue, move || DeviceHandle::event_loop(thread_local, &dev));` so no need for calling `group_clone.enter()`.

neptun/src/device/mod.rs
sfraczek, line 719: Consider hoisting into `new_with_tun` so the mocked path doesn't touch this. This can't work for the mocked socket anyway.

neptun/src/device/mod.rs
stalowyjez: "This assertion about function internals is always true because the function works this way"
I feel like you could write similar comment about almost any unit test xd

TBH most of the function work seems to be ensured by the type system, so the number of the created queues is the most intuitive thing to assert imho

neptun/src/device/mod.rs
stalowyjez: Changed

neptun/src/device/mod.rs
stalowyjez: Sth like this?

neptun/src/device/mod.rs
stalowyjez: Changed

neptun/src/device/mod.rs
stalowyjez: Added a wait

neptun/src/device/mod.rs
stalowyjez: Make sense, fixed

neptun/src/device/mod.rs
stalowyjez: Done

neptun/src/device/mod.rs
stalowyjez: Done

stalowyjez: Or I'd actually leave it without this change, as we wouldn't have to fight with coverage task xd

neptun/src/device/mod.rs
```diff
@@ -263,25 +263,21 @@ impl DeviceHandle {
         let sockets_to_close = Arc::new(Lock::new(vec![]));
         #[cfg(any(target_os = "macos", target_os = "ios", target_os = "tvos"))]
         let threads = {
-            let group = dispatch::Group::create();
-            let mut queues = vec![];
+            let group = dispatch2::DispatchGroup::new();
+            // `global_queue` returns a shared singleton, so fetch it once.
+            let queue =
+                dispatch2::DispatchQueue::global_queue(dispatch2::GlobalQueueIdentifier::Priority(
```
lcruz99: why don't you import dispatch2 objects?

sfraczek (dismissed):

sfraczek:
+1

neptun/src/device/mod.rs
stalowyjez: imported

sfraczek (approved):

PR #74, opened 2026-07-07
Bump crossbeam-epoch dependency

djkarwowski:
Bump `crossbeam-epoch` version in regard to [RUSTSEC-2026-0204](https://rustsec.org/advisories/RUSTSEC-2026-0204).


sfraczek (approved):
+2

PR #75, opened 2026-07-07
LLT-7475: Fix race in event loop registration with GCD group

djkarwowski:
### Problem

This PR addresses two separate issues (it is best to review it commit by commit):
- `read_packet` function implementation did not ensure full compile time safety with regard to a buffer indexed access.
- On Apple platforms, there was a race condition in `DeviceHandle::new()` enabling a short time window, during which an immediate call to `DeviceHandle::wait()` resulted in an instant return. This behavior was breaking `neptun-cli` on macOS. 

### Solution

- The type for the `buf` param for `read_packet` function has been updated to harden the compile time check.
- Ensured atomic registration of the event loop through the `Group::exec_async()` call.

### DIFF
--- a/neptun/src/device/mod.rs
+++ b/neptun/src/device/mod.rs
@@ -1266,11 +1266,12 @@ fn process_iface_inline(
 
 fn read_packet<'a>(
     iface: &Arc<TunSocket>,
-    buf: &'a mut [u8],
+    buf: &'a mut [u8; MAX_PKT_SIZE],
     mtu: &CheckedMtu,
     peers: &AllowedIps<Arc<Peer>>,
 ) -> IfaceReadResult<'a> {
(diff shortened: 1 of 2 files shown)


lcruz99:
does this have ticket #?

lcruz99 (commented):
+1

djkarwowski:
@lcruz99 
> does this have ticket #?

No, but I can create one if that's preferable.

lcruz99:
@djkarwowski it might not be worth for simple and straightforward tasks but I'd say for this bug/issue, it would be good to have a ticket for tracking and context purposes.

stalowyjez:
Lgtm, +1

It has some shared stuff with my dispatch2 PR: https://github.com/NordSecurity/NepTUN/pull/73 - but I guess we'll need to just rebase one of them.

stalowyjez (approved):

PR #76, opened 2026-08-11
LLT-7562, LLT-7644: Rebuild Apple listen sockets on network change and release the old port on listen_port change

sfraczek:
On macOS/iOS/tvOS the WireGuard UDP socket is pinned to a physical interface via `IP_BOUND_IF`, since these platforms have no fwmark equivalent for routing the socket's traffic around the tunnel. When that interface disappears and comes back, the socket's flow control can wedge against the dead incarnation: every send fails with `EAGAIN` and the queue never drains, because the kernel's resume signal can only come from that same incarnation.

`drop_connected_sockets()` is the only thing run on a network change, and it doesn't help - it closes per-peer connected sockets, which are compiled out on Apple.

This rebuilds the shared socket instead: close the old one, then rebind the same port so an existing NAT mapping stays valid when the same interface returns with the same address, falling back to an ephemeral port if that port is taken. Closing has to go through the event loop, since it holds a duplicate of the descriptor that keeps the port allocated.

### DIFF
--- a/neptun/src/device/mod.rs
+++ b/neptun/src/device/mod.rs
@@ -329,6 +329,31 @@ impl DeviceHandle {
 
     pub fn drop_connected_sockets(&self) {
         self.device.read().drop_connected_sockets();
+
+        // No connected sockets on Apple; rebuild the shared socket instead,
+        // or it wedges with EAGAIN after a network change (LLT-7562).
+        #[cfg(any(target_os = "macos", target_os = "ios", target_os = "tvos"))]
+        {
+            let rebuilt = self.device.read().try_writeable(
+                |device| device.trigger_yield(),
+                |device| {
+                    device.cancel_yield();
+                    if device.closed {
+                        return Ok(());
+                    }
+                    device.rebuild_listen_sockets()
+                },
+            );
+            match rebuilt {
+                Some(Ok(())) => {}
+                Some(Err(e)) => {
+                    tracing::error!("Failed to rebuild listen sockets on network change: {e:?}")
+                }
+                None => tracing::error!(
+                    "Failed to gain write access to rebuild listen sockets on network change"
+                ),
+            }
+        }
     }
 
     #[cfg(not(any(target_os = "macos", target_os = "ios", target_os = "tvos")))]
@@ -791,6 +816,29 @@ impl Device {
         self.workers.start(udp4, udp6, &self.config);
 
         self.listen_port = port;
+
+        Ok(())
+    }
+
+    #[cfg(any(target_os = "macos", target_os = "ios", target_os = "tvos"))]
+    fn rebuild_listen_sockets(&mut self) -> Result<(), Error> {
+        // Mirrors wireguard-go's BindUpdate(): close, then rebind the same
+        // port (falls back to a new one) so an existing NAT mapping survives.
+        let port = self.listen_port;
+
+        match self.open_listen_socket(port) {
+            Ok(()) => {
+                tracing::info!("Rebuilt listen sockets on network change, kept port {port}");
+            }
+            Err(e) => {
+                tracing::warn!("Failed to rebind port {port} ({e:?}), taking a new one");
+                self.open_listen_socket(0)?;
+                tracing::info!(
+                    "Rebuilt listen sockets on network change on a new port {}",
+                    self.listen_port
+                );
+            }
+        }
 
         Ok(())
     }
@@ -1485,24 +1533,26 @@ mod tests {
         assert_eq!(checked.get(), mtu);
     }
 
-    #[test]
-    #[cfg(any(target_os = "macos", target_os = "ios", target_os = "tvos"))]
-    fn start_event_loop_threads_starts_one_loop_per_thread() {
-        use std::convert::TryFrom;
+    fn mock_device(n_threads: usize) -> (Device, std::os::unix::net::UnixStream) {
         use std::os::unix::io::IntoRawFd;
         use std::os::unix::net::UnixStream;
-        use std::time::Duration;
 
-        const N_THREADS: usize = 4;
         const MOCK_MTU: usize = 1420;
 
-        // `_far` keeps the socket's peer end open; `near` backs the TunSocket.
-        let (near, _far) = UnixStream::pair().unwrap();
+        // The far end keeps the socket's peer open; `near` backs the TunSocket.
+        // TUNGETIFF would reject a socket pair, so off Apple the fd is wrapped
+        // directly.
+        let (near, far) = UnixStream::pair().unwrap();
+        #[cfg(any(target_os = "macos", target_os = "ios", target_os = "tvos"))]
         let tun = TunSocket::new_from_fd(near.into_raw_fd()).unwrap();
+        #[cfg(not(any(target_os = "macos", target_os = "ios", target_os = "tvos")))]
+        let tun = TunSocket::new_mock_from_fd(near.into_raw_fd());
 
         let config = DeviceConfig {
-            n_threads: N_THREADS,
+            n_threads,
             use_connected_socket: false,
+            #[cfg(target_os = "linux")]
+            use_multi_queue: false,
             open_uapi_socket: false,
             protect: Arc::new(MakeExternalNeptunNoop),
             firewall_process_inbound_callback: None,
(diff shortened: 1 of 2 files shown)


sfraczek:
Follow-up filed: LLT-7623

neptun/tests/device.rs
```diff
@@ -385,6 +385,45 @@ fn test_wireguard_get_on_device() {
...
+    assert_port_held(port);
+}
+
+/// `listen_port=` in the UAPI response is a stored field, so it survives even a
+/// failed rebuild. Binding the port ourselves is what proves a live socket holds
+/// it.
+#[cfg(target_os = "macos")]
+fn assert_port_held(port: u16) {
+    match std::net::UdpSocket::bind(("0.0.0.0", port)) {
+        Ok(_) => panic!("port {} is free, so no socket was rebuilt", port),
```
tomasz-grz: Does this still work after we `set_reuse_address()` (is it `SO_REUSEADDR` not `SO_REUSEPORT` right?).

Although the docs confusingly mention:
>     /// This indicates that further calls to `bind` may allow reuse of local
>     /// addresses. For IPv4 sockets this means that a socket may bind even when
>     /// there's a socket already listening on this port.

neptun/tests/device.rs
```diff
@@ -385,6 +385,45 @@ fn test_wireguard_get_on_device() {
     assert!(response.ends_with("errno=0\n\n"));
 }
 
+#[test]
+#[cfg(target_os = "macos")] // rebuild_listen_sockets is Apple-only
```
tomasz-grz: Should we also include `target_os = "ios", target_os = "tvos"`?

neptun/src/device/mod.rs
```diff
@@ -785,6 +811,72 @@ impl Device {
         Ok(())
     }
 
+    #[cfg(any(target_os = "macos", target_os = "ios", target_os = "tvos"))]
+    fn rebuild_listen_sockets(&mut self) -> Result<(), Error> {
```
tomasz-grz, line 824: A little nitpicky, but there are no sockets directly built in `rebuild_listen_sockets`.
They are rather misleadingly built in `bind_listen_sockets` (maybe would be more accurate to rename to `create_listen_sockets`?)

Which also looks like a lot of code duplication from `open_listen_socket`? Could they be combined to reuse the same bits? Maybe a generic helper to open the socket, set options and do the binding:

```rs
fn create_socket(domain: Domain, port: u16) -> Result<socket2::Socket, Error>
```

neptun/src/device/mod.rs
```diff
@@ -785,6 +811,72 @@ impl Device {
...
+            Err(e) => Err(e),
+        }
+    }
+
+    #[cfg(any(target_os = "macos", target_os = "ios", target_os = "tvos"))]
+    fn bind_listen_sockets(&mut self, mut port: u16) -> Result<(), Error> {
+        let udp_sock4 = socket2::Socket::new(Domain::IPV4, Type::DGRAM, Some(Protocol::UDP))?;
+        udp_sock4.set_reuse_address(true)?;
+        udp_sock4.bind(&SocketAddrV4::new(Ipv4Addr::UNSPECIFIED, port).into())?;
+        udp_sock4.set_nonblocking(true)?;
```
tomasz-grz: Is it ok to `set_nonblocking` and `make_external` after `bind`?

neptun/src/device/mod.rs
```diff
@@ -785,6 +811,72 @@ impl Device {
...
+            port = udp_sock4
+                .local_addr()?
+                .as_socket()
+                .map(|s| s.port())
+                .ok_or_else(|| Error::GetSockName("no local address on new socket".to_owned()))?;
+        }
+
+        let udp_sock6 = socket2::Socket::new(Domain::IPV6, Type::DGRAM, Some(Protocol::UDP))?;
+        udp_sock6.set_reuse_address(true)?;
+        udp_sock6.bind(&SocketAddrV6::new(Ipv6Addr::UNSPECIFIED, port, 0, 0).into())?;
```
tomasz-grz: In case `port` was initially specified `0` and `udp_sock6` fails to bind to the same `port` as `udp_sock4`, should it retry with port `0` as well or do we expect the pair to be always on the same ports?

neptun/tests/device.rs
sfraczek: It did work because the test binds without the flag, and both places would need the option for the reuse to apply.

The flag itself is needed: binding 0.0.0.0:P and then [::]:P fails with EADDRINUSE unless both sockets set it. I added a note to LLT-7623 to document it there.

neptun/src/device/mod.rs
sfraczek: I think it is the standard way, but I am not sure. The same was in open_listen_socket.
The make_external might have to be after bind because on apple make_external->rebind->bind->getsockname
https://github.com/NordSecurity/libtelio/blob/583608721dad9e5f0d214fd1d60b49a71c382da1/crates/telio-sockets/src/protector/apple.rs#L96
and the getsockname needs a bound socket https://docs.rs/nix/latest/nix/sys/socket/fn.getsockname.html

neptun/src/device/mod.rs
sfraczek: Renamed. And I folded the duplication in after all: `rebuild_listen_sockets` now reuses `open_listen_socket`, so `bind_listen_sockets`/`create_listen_sockets` are gone.

neptun/src/device/mod.rs
sfraczek: The pair has to share a port; listen_port is a single value we report and advertise to peers.
The rebuild also never starts with port 0: listen_port always holds the port that was actually bound, even when the device was configured with 0.
If binding the pair fails, rebuild_listen_sockets retries the whole pair on a new port. The case where that retry also fails is written down in LLT-7623.

neptun/tests/device.rs
sfraczek: Done. Also moved the test to the mod.rs test module so CI runs it.

mathiaspeters (commented):
+1

neptun/src/device/mod.rs
tomasz-grz: Makes sense, thanks

neptun/src/device/mod.rs
```diff
@@ -176,6 +176,11 @@ pub struct Device {
     udp4: Option<Arc<socket2::Socket>>,
     udp6: Option<Arc<socket2::Socket>>,
 
+    // The event loop holds a dup of each fd inside its handler closure, so
+    // clearing an event needs that fd, not the original socket's.
+    udp_event_fd4: Option<RawFd>,
+    udp_event_fd6: Option<RawFd>,
```
jjanowsk: I think it would be cleaner to just pass `Arc<socket2::Socket>` to `register_udp_handler` and avoid duplicating the fd entirely. Then you can do this later and  avoid problems with partial cleanup. 
```
self.udp4 = Some(udp4.clone());
self.udp6 = Some(udp6.clone());
 
self.register_udp_handler(udp4.clone())?;
self.register_udp_handler(udp6.clone())?;
```

tomasz-grz (dismissed):
+1 ship it 🚢

jjanowsk (changes requested):

neptun/src/device/mod.rs
```diff
@@ -1520,4 +1578,157 @@ mod tests {
...
+        use std::os::unix::net::UnixStream;
+        use std::time::Duration;
+
+        const MOCK_MTU: usize = 1420;
+
+        // `_far` keeps the socket's peer end open; `near` backs the TunSocket.
+        let (near, _far) = UnixStream::pair().unwrap();
+        let tun = TunSocket::new_from_fd(near.into_raw_fd()).unwrap();
+
+        let config = DeviceConfig {
```
jjanowsk: And can you please extract this to some helper: `fn mock_device(n_threads: usize) -> (Device, std::os::unix::net::UnixStream)`. It's repeated 3 times

neptun/src/device/mod.rs
sfraczek: Done.

neptun/src/device/mod.rs
sfraczek: Done.

matislovas:
+0.5

jjanowsk (approved):

PR #77, opened 2026-08-19
Fix roaming on Apple platforms

jjanowsk:
Recent refactoring of Apple packet processing accidentally broke roaming on those platforms. This fixes it.

### DIFF
--- a/neptun/src/device/mod.rs
+++ b/neptun/src/device/mod.rs
@@ -1142,11 +1142,14 @@ impl Device {
                         }
                     }
 
+                    // Update the peer's endpoint to the source of the last valid packet
+                    // so outbound traffic follows a roaming peer.
+                    peer.set_endpoint(sock);
+
                     #[cfg(not(any(target_os = "macos", target_os = "ios", target_os = "tvos")))]
                     {
                         // This packet was OK, that means we want to create a connected socket for this peer
                         let ip_addr = sock.ip();
-                        peer.set_endpoint(sock);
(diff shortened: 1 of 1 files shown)


neptun/src/device/mod.rs
```diff
@@ -1646,6 +1649,123 @@ mod tests {
...
+        else {
+            panic!("expected keepalive after handshake response");
+        };
+
+        // Roam: send the next valid packet from a different source port.
+        let sock_b = UdpSocket::bind("127.0.0.1:0").unwrap();
+        sock_b.send_to(keepalive, device_addr).unwrap();
+
+        assert!(
+            wait_for_endpoint(sock_b.local_addr().unwrap()),
```
tomaszklak, line 1756: both sock_a and sock_b seems to be defined the same way - will the difference be in the assigned port?

jjanowsk: Yes, that's the point. Without the fix after after  `sock_b.send_to(keepalive, device_addr).unwrap();` device would still hold sock_a's address as an endpoint.

tomaszklak: I'm not sure if it would be possible to do in the test, but having a different IP would be much easier to understand. But treat this as a nitpick :) 

Also, it might be good idea (maybe as a separate ticket) to add such roaming tests to nat-lab in libtelio.

tomaszklak (commented):
+1

djkarwowski (approved):
+2.0

PR #78, opened 2026-08-31
LLT-6820: Add pubkey info to logs

stalowyjez:
# Problem

Currently it's hard to match neptun logs with a peer because there is no peer info attached

# Solution

Add peer field to tracing logs with masked public key

### DIFF
--- a/neptun/src/serialization.rs
+++ b/neptun/src/serialization.rs
@@ -1,6 +1,69 @@
 use base64::{engine::general_purpose, Engine};
 use zeroize::{Zeroize, ZeroizeOnDrop, Zeroizing};
 
+/// Number of leading/trailing base64 characters kept when masking a public key.
+const MASK_KEEP: usize = 4;
+/// base64 length of a 32 byte key: `4 * ceil(32 / 3)`.
+const KEY_BASE64_LEN: usize = 44;
+
+/// Masked public key wrapper.
+///
+/// `Display` and `Debug` both render only the first and last four base64
+/// characters (`AQEB...AQE=`) — enough to tell which peer a log line belongs
+/// to, without putting a whole key in a log sink.
+#[derive(Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub(crate) struct PubKey([u8; 32]);
+
+impl PubKey {
+    /// Raw key material.
+    // Only used by the firewall callbacks; gated to match `device` in lib.rs,
+    // which is unix-only - otherwise this is dead code on Windows.
+    #[cfg(all(unix, feature = "device"))]
+    pub(crate) const fn as_bytes(&self) -> &[u8; 32] {
+        &self.0
+    }
+}
+
+impl From<[u8; 32]> for PubKey {
+    fn from(bytes: [u8; 32]) -> Self {
+        Self(bytes)
+    }
+}
+
+impl From<crate::x25519::PublicKey> for PubKey {
+    fn from(key: crate::x25519::PublicKey) -> Self {
+        Self(key.to_bytes())
+    }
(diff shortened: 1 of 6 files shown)


neptun/src/serialization.rs
```diff
@@ -1,6 +1,84 @@
...
+///
+/// `Display` and `Debug` both render only the first and last four base64
+/// characters (`AQEB...AQE=`) — enough to tell which peer a log line belongs
+/// to, without putting a whole key in a log sink.
+#[derive(Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub(crate) struct PubKey([u8; 32]);
+
+impl PubKey {
+    /// Raw key material.
+    #[allow(dead_code)]
```
tomaszklak: Why is `allow(dead_code)` needed?

neptun/src/serialization.rs
```diff
@@ -1,6 +1,84 @@
 use base64::{engine::general_purpose, Engine};
 use zeroize::{Zeroize, ZeroizeOnDrop, Zeroizing};
 
+/// Number of leading/trailing base64 characters kept when masking a public key.
+const MASK_KEEP: usize = 4;
+/// Separator rendered between the kept fragments of a masked public key.
+const MASK_SEPARATOR: &str = "...";
```
tomaszklak: nit: I don't see the point of putting this into a const

tomaszklak:
Can you attach a sample of new logs?

neptun/src/serialization.rs
```diff
@@ -1,6 +1,84 @@
...
+}
+
+impl From<&crate::x25519::PublicKey> for PubKey {
+    fn from(key: &crate::x25519::PublicKey) -> Self {
+        Self(key.to_bytes())
+    }
+}
+
+impl std::fmt::Display for PubKey {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
```
tomaszklak, line 46: Can you explain why is there so much more code than the old `format_pubkey_short` ?

tomaszklak (commented):
Could we use tracing::Span(::enter) to avoid modifying all the logs explicitly?

stalowyjez:
@tomaszklak I guess it makes ss makes more sense, to capture `tracing` formatting:
image

tomaszklak:
That's from tests, but libtelio has it's own logging setup and I'm curious (and would want ta make sure) how it will look like in production logs.

neptun/src/serialization.rs
stalowyjez: Removed

neptun/src/serialization.rs
stalowyjez: Removed

neptun/src/serialization.rs
stalowyjez: Simplified

neptun/src/serialization.rs
```diff
@@ -1,6 +1,69 @@
 use base64::{engine::general_purpose, Engine};
 use zeroize::{Zeroize, ZeroizeOnDrop, Zeroizing};
 
+/// Number of leading/trailing base64 characters kept when masking a public key.
```
matislovas, line 4: All this serialization stuff is in `telio-crypto`, maybe it is worth to reuse it?

stalowyjez: Well, it's 60 lines of code and tbh pretty verbose one and using telio-crypto would introduce a dependency loop, I don't think it's worth it

matislovas:
+1

tomaszklak (approved):
+1

